PR #6, opened 2019-11-11
[WIP] Denotational Re-Design

FintanH:
Rework the denotational design completely. We revisited the design from 
the beginning, asking what would the user flow be. We came up with a
design that looks like the following:

* A Repo, History, and Snapshot interface
* A Directory view interface
* A Diffing interface

These three components interplay with each other but have clear
boundaries of where one starts and then hands over the information to
the next.

### DIFF
--- a/docs/denotational-design.md
+++ b/docs/denotational-design.md
@@ -254,19 +254,12 @@ getDirectory :: Browser a Directory
 switchHistory :: (History a -> History a) -> Browser a b
 μ switchHistory f = modify f
 
-type ViewResult
-μ ViewResult = Fail | Success
+-- | Find the suffix of a History.
+findSuffix :: Eq a => a -> History a -> Maybe (History a)
+μ findSuffix a = nonEmpty . Nel.dropWhile (/= a)
 
--- View the history up to a given point
--- Returns Success if @a@ existed in the history and we
--- could successfully set a new history.
---
--- Returns Failure if @a@ was not in the history and
--- we did not modify the history.
-viewAt :: Eq a => a -> Browser a ViewResult
-μ viewAt a = do
-  hist <- get
-  case nonEmpty $ Nel.dropWhile (/= a) hist of
-    Nothing -> pure Failure
-    Just h -> setHistory h *> pure Success
+-- View the history up to a given point by supplying a function to modify
+-- the state. If this operation fails, then the default value is used.
+viewAt :: (History a -> Maybe (History a)) -> History a -> Browser a b
+μ viewAt f def = switchHistory (fromMaybe def . f)
 ```
--- a/docs/denotational-design.md
+++ b/docs/denotational-design.md
@@ -66,7 +66,7 @@ type File
 -- Useful for listing a directory and denoting what the label is
 -- corresponding to.
 type SystemType
-SystemType
+μ SystemType
   = IsFile
   | IsDirectory
 
@@ -191,7 +191,7 @@ fuzzyFind :: Label -> [Directory]
 -- A Git Snapshot is grabbing the HEAD commit of your History
 -- and turning it into a Directory
 gitSnapshot :: Snapshot [Commit]
-gitSnapshot = getDirectoryPtr . head
+μ gitSnapshot = getDirectoryPtr . head
 
 -- Opaque and defined by the backend
 getDirectoryPtr :: Commit -> Directory
@@ -199,27 +199,27 @@ getDirectoryPtr :: Commit -> Directory
 -- A Pijul history is semantically applying the patches in a
 -- topological order and achieving the Directory view.
 pijulHistory :: Snapshot Patch
-pijulHistory = foldl pijulMagic mempty
+μ pijulHistory = foldl pijulMagic mempty
 
 -- Opaque and defined by the backend
 pijulMagic :: Patch -> Directory -> Directory
 
 -- Get the current History we are working with.
 getHistory :: Browser a (History a)
-getHistory = get
+μ getHistory = get
 
 -- Get the current Directory in the Browser
 getDirectory :: Browser a Directory
-getDirectory = do
+μ getDirectory = do
   hist <- get
   fromHistory <- ask
   pure $ fromHistory hist
 
 -- We modify the history by changing the internal history state.
 switchHistory :: ([a] -> [a]) -> Browser a b
-switchHistory f = modify f
+μ switchHistory f = modify f
 
 -- View the history up to a given point
 viewAt :: Eq a => a -> Browser a b
-viewAt a = switchHistory (dropWhile (/= a))
+μ viewAt a = switchHistory (dropWhile (/= a))
 ```
--- a/docs/denotational-design.md
+++ b/docs/denotational-design.md
@@ -198,7 +198,7 @@ getDirectoryPtr :: Commit -> Directory
 
 -- A Pijul history is semantically applying the patches in a
 -- topological order and achieving the Directory view.
-pijulHistory :: History Patch
+pijulHistory :: Snapshot Patch
 pijulHistory = foldl pijulMagic mempty
 
 -- Opaque and defined by the backend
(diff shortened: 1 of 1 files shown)


FintanH:
There are a few things that I need to flesh out, but I think this is mostly complete.

docs/denotational-design.md
```diff
@@ -1,214 +1,225 @@
...
+-- Definition elided
+findDirectory :: NonEmpty Label -> Directory -> Maybe Directory
+
+-- Definition elided
+fuzzyFind :: Label -> [Directory]
+
+-- A Git Snapshot is grabbing the HEAD commit of your History
+-- and turning it into a Directory
+gitSnapshot :: Snapshot [Commit]
+gitSnapshot = getDirectoryPtr . head
```
kim: I think this is simply:

```haskell
gitSnapshot :: GitSnapshot -- or: Snapshot Commit
gitSnapshot = ask
```

Although probably not too useful if we have `getDirectory`

docs/denotational-design.md
```diff
@@ -1,214 +1,225 @@
...
+
+-- Opaque and defined by the backend
+getDirectoryPtr :: Commit -> Directory
+
+-- A Pijul history is semantically applying the patches in a
+-- topological order and achieving the Directory view.
+pijulHistory :: History Patch
```
kim:
```suggestion
pijulHistory :: Snapshot Patch
```

docs/denotational-design.md
kim: Actually:

```haskell
gitSnapshot :: Browser GitSnapshot GitSnapshot
```

docs/denotational-design.md
```diff
@@ -1,214 +1,225 @@
...
+-- Get the current History we are working with.
+getHistory :: Browser a (History a)
+getHistory = get
+
+-- Get the current Directory in the Browser
+getDirectory :: Browser a Directory
+getDirectory = do
+  hist <- get
+  fromHistory <- ask
+  pure $ fromHistory hist
```
kim:
```haskell
ask <$> get
```

:eyes:

FintanH: I think I'll keep the version we have for clarity sake :sweat_smile: But maybe change `fromHistory` to `applySnapshot`?

docs/denotational-design.md
FintanH: I was thinking this was more the function to pass to the `ReaderT`, so we would have `runBrowser gitSnapshot $ actions`?

kim: Ah yeah. I think it still doesn't quite check out because how do you take the `head` of `[a] -> b`?

FintanH: Ya, maybe it would be something like:
```haskell
maybe rootDirectory getDirectoryPtr $ Nel.head <$> nonEmpty
```

Or even saying that! We should probably have `NonEmpty a` as our set of artefacts. `Repo` should be `Maybe (NonEmpty (History a))`?

FintanH: That way the only we have `Nothing` is when our `Repo` is truly in the initial state of nothingness.

docs/denotational-design.md
```diff
@@ -1,214 +1,228 @@
...
+gitSnapshot :: Snapshot [Commit]
+μ gitSnapshot = getDirectoryPtr . head
+
+-- Opaque and defined by the backend
+getDirectoryPtr :: Commit -> Directory
+
+-- A Pijul history is semantically applying the patches in a
+-- topological order and achieving the Directory view.
+pijulHistory :: Snapshot Patch
+μ pijulHistory = foldl pijulMagic mempty
```
FintanH: @kim Actually this seems to be broken as well. We would actually need to do `foldl pijulMagic rootDir` where `rootDir :: Directory`, but that's not possible since what do we put in the `NonEmpty`? :thinking:

docs/denotational-design.md
FintanH: Actually `Maybe (NonEmpty (History a))` is the same as `[History a]` :sweat_smile:

docs/denotational-design.md
FintanH: Unless we have:
```haskell
rootDir :: Directory
rootDir = ("~", Right (".root", empty) :| [])
```
i.e. the root directory is the root label with a root file.
```

Comment (FintanH) [reply docs/denotational-design.md]: It feels a bit hacky though... But we do need some idea of a `root` at the same time so that we can begin a traversal.

Comment (kim) [reply docs/denotational-design.md]:
```haskell
gitRoot = ("~", Left (".git", ...) :| [])
pijulRoot = ("~", Left (".pijul", ...) :| [])
```

is accurately describing reality ;)

Except that those directories are magical and should not be browseable -- does that need to be captured in the semantics?

FintanH: But what's the first file we tie into those? :thinking: Or maybe this is some magic opaque trait that is `getRoot :: Directory`.

kim: How about something like this:

```haskell
type DirectoryContents
μ DirectoryContents = IsRepo | Directory | File

-- Opaque representation of repository state directories (e.g. `.git`, `.pijul`)
-- Those are not browseable, but have to be present at the repo root 'Directory'.
type IsRepo

type Directory
μ Directory = (Label, NonEmpty DirectoryContents)

emptyRepoRoot :: Directory
μ emptyRepoRoot = ("~", IsRepo :| [])
```

FintanH: Oh that's cool! Because we can have sub vcs directories inside directories!

FintanH: One downside of this is that it's possible to represent:
```haskell
wut :: Directory
wut = ("wut", IsRepo :| [IsRepo, IsRepo, IsRepo, IsRepo])
```
But we can hide those details from the user I guess :thinking:

kim: Hm. I'm sure this can be fixed somehow using `absurd` 🤣

FintanH: Or just regular smart constructors :eyes: :sweat_smile: Going forward with this design :+1:

docs/denotational-design.md
```diff
@@ -1,214 +1,260 @@
...
+  applySnapshot <- ask
+  pure $ applySnapshot hist
+
+-- We modify the history by changing the internal history state.
+switchHistory :: ([a] -> [a]) -> Browser a b
+μ switchHistory f = modify f
+
+-- View the history up to a given point
+viewAt :: Eq a => a -> Browser a b
+μ viewAt a = switchHistory (dropWhile (/= a))
```
FintanH: Oh no... I missed this when changing `History` to `NonEmpty`. This doesn't quite work. Need something like `Maybe (Browser a b)` because we can't be working over an empty history if we can't find the artifact.

FintanH: But that type sig doesn't seem to make sense... maybe something like:
```haskell
type ViewResult
  = Fail
  | Succes

viewAt :: Eq a => a -> Browser a ViewResult
μ viewAt a = do
  hist <- get
  let new_hist = nonEmpty $ Nel.dropWhile (/= a) hist
  case new_hist of
    Nothing -> pure Failure
    Just h -> setHistory h *> pure Success
```

FintanH: @kim Any thoughts on making this nicer?

kim: Hm, yeah, it's not entirely satisfying, because the caller needs to scrutinise the monadic result, while the state stays the same. The honest thing to do would be to diverge (`throw`), or add an `ExceptT` layer.

Or, let the caller decide which `History` to default to:

```haskell
findSuffix :: Eq a => a -> History a -> Maybe (History a)
μ findSuffix a = nonEmpty . Nel.dropWhile (/= a)

viewAt :: (History a -> Maybe (History a)) -> History a -> Browser a b
μ viewAt f def = switchHistory (fromMaybe def . f)
```

FintanH: Ya that seems cleaner!

PR #7, opened 2019-11-13
New Design Implementation

FintanH:
This is a follow up of #6 that implements the Rust bindings for the denotational spec


src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+            search_directory
+        } else {
+            None
+        }
+    }
+
+    // TODO(fintan): This is going to be a bit trickier so going to leave it out for now
+    #[allow(dead_code)]
+    pub fn fuzzy_find(_label: Label) -> Vec<Self> {
+        panic!("TODO")
```
cloudhead: There's an `unimplemented!()` macro FYI

FintanH: TIL! Thank you :)

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+pub trait RepoBackend
+where
+    Self: Sized,
+{
+    fn new() -> Directory<Self>;
+}
+
+#[derive(Debug, Clone)]
+pub enum DirectoryContents<Repo> {
+    SubDirectory(Box<Directory<Repo>>),
```
cloudhead: Does this need to be boxed?

cloudhead: Ah is it because of the recursion?

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum SystemType {
+    IsFile,
+    IsDirectory,
+}
+
+impl SystemType {
+    pub fn is_file(label: Label) -> (Label, Self) {
+        (label, SystemType::IsFile)
+    }
```
cloudhead: Usually `is_*` functions in rust are used for returning `bool`, so could be confusing. Maybe change the enum to `File, Directory`, and then just call this `file`? :man_shrugging:

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    pub fn find_directory(&self, path: Path) -> Option<Self>
+    where
+        Repo: Clone,
+    {
+        let mut search_directory = Some(self.clone());
+        let (label, labels) = path.split_first();
+        if *label == self.label {
+            for label in labels {
+                match search_directory {
```
cloudhead: Something about this is funky - we know the value of `search_directory` before we start the loop, but we're checking it inside the loop. :thinking:

cloudhead: I guess it's the `None` branch that seems out of place, because we're not using the `label` at all to determine whether to take that branch.

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    fn file_in_directory(&self, label: &Label) -> Option<File> {
+        for entry in self.entries.iter() {
+            match entry {
+                DirectoryContents::File(file) => {
+                    if file.filename == *label {
+                        return Some(file.clone());
+                    } else {
+                        continue;
+                    }
```
cloudhead: You can if you want do `DirectoryContents::File(file) if file.filename == *label { return ... }`

cloudhead: But you'll need a catch-all afterwards. Still, I think it would be cleaner.

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+                }
+                DirectoryContents::SubDirectory(_) => continue,
+                DirectoryContents::Repo(_) => continue,
```
cloudhead: You probably know this, but you can just do `DirectoryContents::Repo(_) => {},`

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+#[derive(Debug, Clone)]
+pub struct Directory<Repo> {
+    pub label: Label,
+    pub entries: NonEmpty<DirectoryContents<Repo>>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct File {
+    pub filename: Label,
+    pub contents: String,
```
cloudhead: No binary data support? I'd expect `Vec<u8>` here, but you know what you need better than I do :)

src/file_system/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+impl From<String> for Label {
+    fn from(item: String) -> Self {
+        Label(item)
+    }
+}
+
+pub struct Path(pub NonEmpty<Label>);
```
cloudhead: I guess you've looked at `std::path::PathBuf` already?

src/vcs/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn get_history(identifier: &Self::HistoryId, repo: Repo<A>) -> History<A>;
+
+    fn get_identifier(artifact: &A) -> &Self::ArtefactId;
+
+    fn find_in_history(identifier: &Self::ArtefactId, history: History<A>) -> Option<A>
+    where
+        A: Clone,
+        Self::ArtefactId: PartialEq,
+    {
+        let history: Vec<A> = history.0.into();
```
cloudhead: Do you need to convert it to a `Vec` to iterate? It will be a full copy which may be expensive on large histories.

src/file_system/mod.rs
FintanH: Yup! Basically a `Directory` is `NonEmpty` rose-tree

src/file_system/mod.rs
FintanH: Ya, good point! I forgot the scoping is better in Rust over Haskell :sweat_smile:

src/file_system/mod.rs
FintanH: I have not :eyes: I'll stick it on my todo list to check it out. Might be useful as a back-end for this implementation :thinking:

FintanH: One thing that we want to keep as an invariant is that `Directory` should not be empty, since an empty directory does not exist in VCSes

src/file_system/mod.rs
FintanH: Tbh, in the design doc it is a `ByteString` but I didn't want to commit to something that's harder to test than `String` for now. It'll be fleshed out once I get to the stage of writing `diff` :)

src/file_system/mod.rs
FintanH: Ya, so it's a search over the `labels`. We keep digging down into sub-directories, updating `search_directory` each time. If we don't find it then we bail out with `return None`, otherwise we keep going. If we reach the end of the labels, we'll either have found our directory or not. 

1) Would it help to add comments?
2) Can you see a cleaner way of writing it? :grimacing:

src/file_system/mod.rs
FintanH: I did not know this. I wonder what's clearer though :thinking:

src/file_system/mod.rs
FintanH: Is that the same as Python's if comprehension? It looks kind of weird to me. Can you point me to something explaining it?

src/vcs/mod.rs
FintanH: I don't, thanks for catching that! It was a temporary solution for when `iter` on `NonEmpty` had the type inference issues :)

src/file_system/mod.rs
kim: Like a pattern guard in Haskell, no? Methinks you could just write:

```rust
match entry {
    DirectoryContents::File(file) if file.filename == *label => return Some(file.clone());
    _ => continue;
}
```

src/file_system/mod.rs
kim: Maybe `try_fold` over the tail of the `path`, until `get_sub_directory` returns `None`?

src/file_system/mod.rs
cloudhead: It's like a guard in haskell :grin:

cloudhead: https://doc.rust-lang.org/1.1.0/book/patterns.html#guards

src/file_system/mod.rs
cloudhead: well, to me, since `continue` is the default, you don't need to spell it out, unless there's code after the `continue` that you want to skip over. But here it would continue regardless.

src/file_system/mod.rs
cloudhead: But this is a path, not a directory? Or am I misunderstanding?

kim: Yeah true, the `Directory` is already non-empty :thinking:

FintanH: `Path` is the way to specify where the user will be searching in a `Directory` though

src/file_system/mod.rs
FintanH: I'll take a look.

@kim I like laying out the cases to ensure there's no fall through when adding a case.

src/file_system/mod.rs
FintanH: So if we're searching a directory there's no point in supplying an empty path

src/vcs/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    pub fn modify_history<F>(&mut self, f: F)
+    where
+        F: Fn(&History<A>) -> History<A>,
+    {
+        self.history = f(&self.history)
+    }
+
+    pub fn view_at(&mut self, artifact: A) -> ViewResult
```
FintanH: TODO: Change to use new semantics

src/file_system/mod.rs
FintanH: Created an issue: https://github.com/radicle-dev/radicle-surf/issues/8

src/file_system/mod.rs
FintanH: Created an issue for this https://github.com/radicle-dev/radicle-surf/issues/9 so we don't forget

PR #15, opened 2019-11-25
Add size function for File and Directory

### LINKED ISSUE #12: Implement `size` for `Directory`
See: https://github.com/oscoin/mvp/pull/55/files#r349635280

FintanH:
Fixes #12 

Can get the `size` of a `File` and a `Directory`

### DIFF
--- a/src/file_system/mod.rs
+++ b/src/file_system/mod.rs
@@ -832,7 +832,7 @@ pub mod tests {
         let mut directory_map = HashMap::new();
 
         // Root files set up
-        let mut root_files = (
+        let root_files = (
             File::new("foo.rs".into(), b"use crate::bar"),
             vec![File::new("bar.rs".into(), b"fn hello_world()")],
         )
@@ -840,7 +840,7 @@ pub mod tests {
         directory_map.insert(Path::root(), root_files);
 
         // Haskell files set up
-        let mut haskell_files = (
+        let haskell_files = (
             File::new("foo.hs".into(), b"module Foo where"),
             vec![File::new("bar.hs".into(), b"module Bar where")],
         )
--- a/src/file_system/mod.rs
+++ b/src/file_system/mod.rs
@@ -424,6 +424,27 @@ pub struct File {
     pub contents: Vec<u8>,
 }
 
+impl File {
+    /// Get the size of the `File` corresponding to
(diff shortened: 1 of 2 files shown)


src/file_system/mod.rs
```diff
@@ -473,6 +494,64 @@ impl Directory {
         Directory::mkdir(Label::root(), Repo::repo_directory())
     }
 
+    /// Get the total size, in bytes, of a `Directory`. The size is
+    /// the sum of all files that can be reached from this `Directory`.
```
xla: This implies recursively getting the size for all sub-directories, is that true? How expensive is this operation? Thinking esp. about not so modern storage devices and/or congested user machines.

src/file_system/mod.rs
```diff
@@ -473,6 +494,64 @@ impl Directory {
...
+    /// entries.push(subdir);
+    ///
+    /// let directory = Directory {
+    ///     label: Label::root(),
+    ///     entries: entries,
+    /// };
+    ///
+    /// assert_eq!(directory.size(), 66);
+    /// ```
+    pub fn size(&self) -> usize {
```
kim: Hm, this can get very expensive. Do you think the `File` abstraction could be made such that it reads in the file contents lazily and just peeks at `blob.size()`?

FintanH: Just to clarify, do you mean it would be expensive to compute?

One thing I currently have is that I can grab the size from the `Blob` itself. I can push that commit in a second :)

kim: Yeah, you're reading the contents of all blobs into memory. It seems like [Blob::size()](https://docs.rs/git2/0.10.2/src/git2/blob.rs.html#38) and [Blob::content()](https://docs.rs/git2/0.10.2/src/git2/blob.rs.html#29) access the filesystem only when requested, so potentially `File` could delegate to the backend.

Probably good for now, but could use an optimisation pass later.

FintanH: Ah ok, so have `contents` and `size` be upon request rather than hold the data immediately. I'll create an issue for that and at least Application have the info to work with for the time being :)

FintanH: https://github.com/radicle-dev/radicle-surf/issues/17

PR #21, opened 2019-11-29
Replace quickcheck with proptest

FintanH:


src/file_system/mod.rs
```diff
@@ -894,14 +868,46 @@ pub mod tests {
...
 
-    /* TODO(fintan): this quickcheck takes far too long to complete
-    #[quickcheck]
-    fn prop_all_directories_and_files_quickcheck(
-        directory_map: SmallHashMap<Path, (File, Vec<File>)>,
-    ) -> bool {
-        prop_all_directories_and_files(directory_map.get_small_hashmap)
+    fn label_strategy() -> impl Strategy<Value = Label> {
+        // ASCII regex, see: https://catonmat.net/my-favorite-regex
+        "[ -~]*".prop_map(|label| label.into())
```
kim: That's a cool regex, and I want the T-shirt!

Buuuut, it contains `/`, which I think is not quite correct here.

FintanH: Ah, good catch!

> I want the T-shirt!

I know what to get you for Xmas

xla: > I know what to get you for Xmas

A `/`?

Issue #22, opened 2019-11-29
Implement diff


FintanH:
Currently the diff'ing of `Directory` is unimplemented. This needs to be fleshed out.

FintanH:
Closing because there are enough diff'ing tickets with more prose

PR #24, opened 2019-12-02
Implement diff #22

atomashpolskiy:
Hey guys,

Please let me know, if this contribution would be welcome, and if yes, I will add automatic tests (and update the PR, if you have any objections/suggestions).

You may run the example with the following command (from the project's root directory):

```
cargo run --release --example diff . \
e88d1d8e34212f2dfa9d34d2d2005932fd84cb06 HEAD
```

It's possible to try it out with a different repository, e.g.:

```
cargo run --release --example diff /Users/sadpotato/git-workspace/cassandra/ \
f31c53049d37b16e413eb9ae7d481014f5391e9c \
1c41a9ac2c147ed111d9d8fba53652707dac7df0
```

You may also check out [here](https://github.com/atomashpolskiy/radicle-surf/pull/1) for my ramblings on the issue.


atomashpolskiy:
@FintanH , thanks for taking the time to review! Will look into all of your points shortly.

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
...
+        let mut new_files: Vec<DeleteFile> = Diff::collect_files_from_entry(
+            dc, &parent_path, Diff::convert_to_deleted)?;
+        self.deleted.append(&mut new_files);
+        Ok(())
+    }
+
+    fn build_non_empty_path(file: &File, parent_path: &Rc<RefCell<Vec<Label>>>) -> NonEmpty<Label> {
+        let mut path = parent_path.borrow().to_vec();
+        path.push(file.filename.clone());
+        // path is always non-empty, so we can use unwrap()
```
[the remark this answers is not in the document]
atomashpolskiy: Right above this comment we insert an element into the path, so it can't be empty.

.gitignore
```diff
@@ -8,3 +8,5 @@ Cargo.lock
 
 # These are backup files generated by rustfmt
 **/*.rs.bk
+.idea
+*.iml
```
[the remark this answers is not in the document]
atomashpolskiy: Done

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
+#![allow(dead_code, unused_variables)]
+
 use crate::file_system::{Directory, Label};
 use nonempty::NonEmpty;
+use crate::file_system::{DirectoryContents, File};
+use std::cmp::Ordering;
+use std::cell::RefCell;
+use std::rc::Rc;
+
+type Result = std::result::Result<Diff, DiffError>;
```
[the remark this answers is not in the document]
atomashpolskiy: Done

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
...
+        // Final decision can be based on heuristics, e.g. the file can be considered moved,
+        // if len(LCS) > 0,25 * min(size(d), size(c)), and deleted otherwise.
+
+        Ok(diff)
+    }
+
+    fn collect_diff(old: &Directory, new: &Directory, parent_path: &Rc<RefCell<Vec<Label>>>, diff: &mut Diff)
+        -> std::result::Result<(), String> {
+
+        // TODO: Consider storing directory contents in sorted order
```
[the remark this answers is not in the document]
atomashpolskiy: Guess I'm keeping the TODO for now then?

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
...
+                    (DirectoryContents::SubDirectory(new_dir), DirectoryContents::File(old_file)) => {
+                        diff.add_created_files(new_entry.unwrap(), parent_path)?;
+                        diff.add_deleted_file(old_file, parent_path);
+                    },
+                    (DirectoryContents::SubDirectory(new_dir), DirectoryContents::SubDirectory(old_dir)) => {
+                        parent_path.borrow_mut().push(new_dir.label.clone());
+                        Diff::collect_diff(&**old_dir, &**new_dir, parent_path, diff)?;
+                        parent_path.borrow_mut().pop();
+                    },
+                    _ => panic!("should not happen unless the algo is incorrect")
```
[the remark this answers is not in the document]
atomashpolskiy: Me neither, but can't come up with a better alternative. Adding match arms for `(Repo, _)` and `(_, Repo)` would complicate things quite a bit...

src/diff/mod.rs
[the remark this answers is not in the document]
FintanH: Ya, I created #25 and this can be revisited :)

src/diff/mod.rs
```diff
@@ -1,36 +1,278 @@
...
+                                (SubDirectory(new_dir), File(old_file)) => {
+                                    diff.add_created_files(new_entry, parent_path)?;
+                                    diff.add_deleted_file(old_file, parent_path);
+                                },
+                                (SubDirectory(new_dir), SubDirectory(old_dir)) => {
+                                    parent_path.borrow_mut().push(new_dir.label.clone());
+                                    Diff::collect_diff(&**old_dir, &**new_dir, parent_path, diff)?;
+                                    parent_path.borrow_mut().pop();
+                                },
+                                _ => panic!("should not happen unless the algo is incorrect")
```
FintanH: Can we add the check for `Repo` here instead of above? This would avoid the panic and read well for covering all cases. Also it avoids the fact that a case might be added and then suddenly things panic by accident.

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
...
+            DirectoryContents::SubDirectory(subdir) => subdir.label.clone(),
+            DirectoryContents::File(file) => file.filename.clone(),
+            DirectoryContents::Repo => Label::from(""),
+        }
+    });
+    vec
+}
+
+// TODO: Move somewhere?
+fn into_non_empty<T>(vec: &Vec<T>) -> std::result::Result<NonEmpty<T>, String> where T: Clone {
```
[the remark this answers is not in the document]
atomashpolskiy: Done

src/diff/mod.rs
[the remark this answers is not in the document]
atomashpolskiy: Nevermind, I think I've got it. Check out the new version.

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
...
+}
+
+// returns list of contents, sorted by label; Repos are prepended to the beginning
+fn get_sorted_contents(dir: &Directory) -> Vec<&DirectoryContents> {
+    let mut vec: Vec<&DirectoryContents> = dir.entries.iter().collect();
+    vec.sort_by_key(|e| {
+        match e {
+            DirectoryContents::SubDirectory(subdir) => subdir.label.clone(),
+            DirectoryContents::File(file) => file.filename.clone(),
+            DirectoryContents::Repo => Label::from(""),
```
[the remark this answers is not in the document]
atomashpolskiy: Nice trick!

src/diff/mod.rs
```diff
@@ -1,36 +1,285 @@
...
+        let mut old_iter = old.iter();
+        let mut new_iter = new.iter();
+        let mut old_entry_opt = old_iter.next();
+        let mut new_entry_opt = new_iter.next();
+
+        while old_entry_opt.is_some() || new_entry_opt.is_some() {
+            match (old_entry_opt, new_entry_opt) {
+                (Some(old_entry), Some(new_entry)) => {
+                    let cmp = new_entry.label().cmp(&old_entry.label());
+                    match cmp {
```
FintanH: This can just be `match new_entry.label().cmp(&old_entry.label())` :)

src/diff/mod.rs
```diff
@@ -1,36 +1,285 @@
...
-        path: NonEmpty<Label>,
-    },
-    DeleteFile {
-        path: NonEmpty<Label>,
-    },
-}
-
-pub struct Diff(pub Vec<Change>);
+pub struct CreateFile {
+    pub path: NonEmpty<Label>,
```
FintanH: Actually, why aren't these `Path`s?

src/diff/mod.rs
```diff
@@ -1,36 +1,285 @@
...
+                                    new_entry_opt = new_iter.next();
+                                },
+                                (SubDirectory(new_dir), SubDirectory(old_dir)) => {
+                                    parent_path.borrow_mut().push(new_dir.label.clone());
+                                    Diff::collect_diff(&**old_dir, &**new_dir, parent_path, diff)?;
+                                    parent_path.borrow_mut().pop();
+                                    old_entry_opt = old_iter.next();
+                                    new_entry_opt = new_iter.next();
+                                },
+                                (_, _) => {
```
FintanH: I'd rather have an explicit match here and we just call `next` and let the loop do its thing. Realistically, we'll usually only see one `Repo` and that's at the root.

src/diff/mod.rs
```diff
@@ -1,36 +1,281 @@
...
+                    files.push(mapper(file, path)?);
+                },
+                DirectoryContents::Repo => { /* Skip repo directory */ }
+            }
+        }
+        parent_path.borrow_mut().pop();
+        Ok(())
+    }
+
+    fn convert_to_deleted(file: &File, path: Vec<Label>) -> std::result::Result<DeleteFile, String> {
```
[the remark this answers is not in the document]
atomashpolskiy: Switched to using Path

src/diff/mod.rs
atomashpolskiy: Outdated, will respond in the newer comment

src/diff/mod.rs
atomashpolskiy: Done

src/diff/mod.rs
atomashpolskiy: Switched to using Path and simplified some other things along the way

src/diff/mod.rs
atomashpolskiy: Added explicit matches for Repo combinations.

src/diff/mod.rs
atomashpolskiy: Seems like by doing so I screwed up the order of path elements (they are reversed now). Need to look into the API closer..

FintanH: `Path`s are as you would read them. `foo/bar/baz.rs` is `(foo, vec![bar, baz.rs])`

src/file_system/mod.rs
```diff
@@ -397,6 +399,14 @@ impl DirectoryContents {
     pub fn file(filename: Label, contents: &[u8]) -> Self {
         DirectoryContents::File(File::new(filename, contents))
     }
+
+    pub fn label(&self) -> Option<&Label> {
+        match self {
+            DirectoryContents::SubDirectory(dir) => Some(&dir.label),
+            DirectoryContents::File(file) => Some(&file.filename),
+            _ => None
```
FintanH: Explicit match here, please. In the case of the `enum` getting extended we'll get a compiler warning

src/diff/mod.rs
```diff
@@ -1,36 +1,269 @@
...
+            moved: Vec::new(),
+            modified: Vec::new(),
+        }
+    }
+
+    // TODO: Direction of comparison is not obvious with this signature.
+    // For now using conventional approach with the right being "newer".
+    pub fn diff(left: Directory, right: Directory) -> Result<Diff, DiffError> {
+        let mut diff = Diff::new();
+        let path = Rc::new(RefCell::new(Vec::new()));
```
FintanH: Can you should some light on the use of this `RefCell` actually? I can't seem to grok its use being threaded through. And what do you think about using a `Path` for it? In my suggestion it was `Path::root()` but it can be the shared label of `left` and `right`. If they're not shared then the `Diff` would either be a `mv` or just a completely different directory :thinking:

src/diff/mod.rs
```diff
@@ -1,36 +1,269 @@
...
+            path: Path::from_labels(file.filename.clone(), parent_path)
+        }
+    }
+
+    fn add_modified_file(&mut self, file: &File, parent_path: &Vec<Label>) {
+        // TODO: file diff can be calculated at this point
+        // Use pijul's transaction diff as an inspiration?
+        // https://nest.pijul.com/pijul_org/pijul:master/1468b7281a6f3785e9#anesp4Qdq3V
+        self.modified.push(ModifiedFile {
+            path: Path::from_labels(file.filename.clone(), parent_path),
```
FintanH: I think you want `parent_path.push(file.filename.clone())`, that's why you're getting the reversal

src/diff/mod.rs
```diff
@@ -1,36 +1,269 @@
...
+        -> Result<Vec<T>, String>
+        where F: Fn(&File, &Vec<Label>) -> T + Copy {
+
+        match entry {
+            DirectoryContents::SubDirectory(dir) => Diff::collect_files(dir, parent_path, mapper),
+            DirectoryContents::File(file) => {
+                let mapped = mapper(file, &RefCell::borrow(parent_path));
+                Ok(vec![mapped])
+            },
+            _ => Err(String::from("Unexpected entry type"))
```
FintanH: I don't think it's unreasonable to do `DirectoryContents::Repo => vec![]` here

src/diff/mod.rs
```diff
@@ -247,6 +247,19 @@ impl Diff {
         self.deleted.append(&mut new_files);
         Ok(())
     }
+
+    fn build_path(file: &File, parent_path: &Vec<Label>) -> Path {
+        match parent_path.len() {
+            0 => {
+                // TODO: Path::from_label ?
+                Path::from_labels(file.filename.clone(), &[])
```
FintanH: I believe this would imply that we're at a root folder? So you could do `Path::with_root(&[file.filename.clone()])`

src/diff/mod.rs
```diff
@@ -247,6 +247,19 @@ impl Diff {
         self.deleted.append(&mut new_files);
         Ok(())
     }
+
+    fn build_path(file: &File, parent_path: &Vec<Label>) -> Path {
+        match parent_path.len() {
```
FintanH: `if parent_path.is_empty()`

src/diff/mod.rs
```diff
@@ -247,6 +247,19 @@ impl Diff {
...
+            0 => {
+                // TODO: Path::from_label ?
+                Path::from_labels(file.filename.clone(), &[])
+            },
+            _ => {
+                Path::from_labels(parent_path[0].clone(),
```
FintanH: Hmmmm this is making me think we should be passing around `Path` vs `Vec<Label>` even more :thinking:

FintanH:
If you'd like to fix the test suite you just need to go to `lib.rs` and add `SystemType::directory("examples")` before the `"src"` entry :)

src/diff/mod.rs
atomashpolskiy: > Can you should some light on the use of this RefCell actually? I can't seem to grok its use being threaded through. And what do you think about using a Path for it?

We need RefCell to be able to modify the path vector during the recursive descent into a subdirectory (in `collect_files_inner`). It would be nice to use a Path object, but it does not support removal of path elements.

> In my suggestion it was Path::root() but it can be the shared label of left and right.

It is exactly the shared label (path, more precisely) of some directory, that is present in both revisions.

> If they're not shared then the Diff would either be a mv or just a completely different directory

Correct, that's how the algorithm works. It traverses the file tree in both revisions (typename Directory may be confusing in this context) in parallel collecting changes along the way. At some point the revisions diverge: some directory is present only in one of the revisions. We exploit this fact to quickly collect all files in this directory and transform them into changes, by the following rule:

- if the directory is present in left, but not present in right, then it and all of its' files have been deleted
- if the directory is present in right, but not present in left, then it and all of its' files have been created

src/diff/mod.rs
atomashpolskiy: I'm afraid this change is in 7fa185957e6ba1d82dd1d37d0c7c3a2fe04f2634 which is not yet merged in master

src/file_system/mod.rs
atomashpolskiy: Yup, makes total sense

src/diff/mod.rs
atomashpolskiy: It's not unreasonable, but why? When (and if) the Repo type is possibly updated to reflect embedded repositories (like a git subtree, for instance), returning an empty vec for it will become a programming error. Why not fail early and increase the changes to find this in tests?

src/diff/mod.rs
FintanH: Hmm you may want to `git fetch`, it's there: https://github.com/radicle-dev/radicle-surf/blob/master/src/file_system/mod.rs#L271

src/diff/mod.rs
atomashpolskiy: Not sure if serious? :)

src/diff/mod.rs
atomashpolskiy: Yup, missed it, thanks

src/diff/mod.rs
FintanH: Actually, I believe that would be accounted for in this design :) It would be a `SubDirectory` where one of its entries is a `Repo` :)

src/diff/mod.rs
atomashpolskiy: I suggest to discuss the issue of modifying Path in the comment above

src/diff/mod.rs
atomashpolskiy: What about the name? For instance, Git subtree is connected to a directory, so it needs a label in your model.

src/diff/mod.rs
FintanH: Hmmm, so if I'm understanding this correctly, maybe we should choose the `right` directory as the root path to work from? Since it's the latest revision the `left` should either have the same root or everything got moved/deleted.

I expect that the majority of the time the case will be that `left` and `right` both work from `Path::root` anyway :)

src/diff/mod.rs
FintanH: :+1:

src/diff/mod.rs
FintanH: Do you mean the `.git` directory?

For all intents and purposes of browsing a VCS file system we won't want to reveal what's in `.git` since it's treated as a special directory. For example, you don't see those folders in GitHub or Nest.

src/diff/mod.rs
atomashpolskiy: > Hmmm, so if I'm understanding this correctly, maybe we should choose the right directory as the root path to work from? Since it's the latest revision the left should either have the same root or everything got moved/deleted.

As I see it, we don't need to discern between root and non-root paths, as long as we traverse the directories in parallel. To put it differently, we don't need a point of reference in this algorithm. The differences between left and right stand out naturally as it is.

> I expect that the majority of the time the case will be that left and right both work from Path::root anyway :)

Why? E.g. I'm able to select a directory in IDE and click "Git -> Show history". It will display only the changes for this particular directory.

FintanH: > As I see it, we don't need to discern between root and non-root paths, as long as we traverse the directories in parallel. To put it differently, we don't need a point of reference in this algorithm. The differences between left and right stand out naturally as it is.

I think I see your point, I just think it might make it simpler when writing the algorithm. But at the same time, I still need to grok the algorithm a bit so I could be wrong! :) Maybe what I'll try to do is write some tests for and I can explore how it works, do you think that would help both of us?

> Why? E.g. I'm able to select a directory in IDE and click "Git -> Show history". It will display only the changes for this particular directory.

Sure but that would likely be done by:
```rust
// browser set up already done
let left = browser.get_directory();
browser.commit("...sha1");
let right = browser.get_directory();
diff = Diff::diff(left, right);
```

Does that make sense?

atomashpolskiy: > Maybe what I'll try to do is write some tests for and I can explore how it works, do you think that would help both of us?

Absolutely, if you have the time and desire to do so :)

> Sure but that would likely be done by:

I think that in this case you will be comparing the entire repository between two revisions. At least that's how it worked for me in the [runnable example](https://github.com/radicle-dev/radicle-surf/pull/24/files#diff-d2aa9f7ebc89743fb95e0a3db3b9e42d).

src/diff/mod.rs
atomashpolskiy: Ah, so Repo is for `.git`? Sorry, I was thinking it means something else entirely. In this case your initial comment makes total sense.

src/diff/mod.rs
FintanH: > Absolutely, if you have the time and desire to do so :)

Ya, this was next on my list of things to do anyway! So again, I really appreciate the upfront work and it'd be my pleasure to add some test coverage :) I'll get around to this tomorrow and can make a PR against yours.

> I think that in this case you will be comparing the entire repository between two revisions. At least that's how it worked for me in the runnable example.

I _believe_ that's what I'm describing in my pseudo-code. A browser holds onto a repository and in the snippet we're looking at a particular commit and looking at its snapshot (the directory at that point in time).

I'll play around with your CLI as well :)

atomashpolskiy: > I believe that's what I'm describing in my pseudo-code. A browser holds onto a repository and in the snippet we're looking at a particular commit and looking at its snapshot (the directory at that point in time).

Yes, but I was talking about viewing history for an actual directory (folder), not the repository:

src/diff/mod.rs
FintanH: Ah my bad! I should have clarified sooner :grin: We got there in the end!

src/diff/mod.rs
FintanH: They're essentially linked. The point of this is that to get a `Directory` you need to materialise it using a `History`. `Browser` encapsulates this by having a `Snapshot` function `History -> Directory` and a current `History` state it's looking at.

Does that make sense?

atomashpolskiy: Yes, but `get_directory()` returns root, right? So you can't get the `/src/diff` directory with this method? You need to navigate via two calls to `Directory.sub_directory()`. In which case, if you pass the resulting Directory to `diff()`, it will not work from `Path::root`, but from `/src/diff`. So, assuming there's only a single file in this directory with name `mod.rs` and it's been modified, then the resulting Diff's `modified` list will contain a single element:

```rust
ModifiedFile {
  path: Path(NonEmpty(Label("mod.rs"), Nil))
}
```

atomashpolskiy: Huh, I just noticed that `with_root()` method prepends `~` label to the path. Which is incorrect, given what I'm talking about above.

FintanH: `get_directory` returns the whole directory tree _from_ the root. So if we _were_ to use `get_directory` on two different browser states the `~` will match and then it well recursively check from there. In this case it would check `src`, then `diff`, then do the file diff on `mod.rs`.

But I'm not saying that we should assume that we always start from the root though, I'm just saying that it will be the most common case :)

FintanH:
So I started adding some test on this branch https://github.com/radicle-dev/radicle-surf/tree/fintan/add-diff-tests.

So I added a failing test here: https://github.com/radicle-dev/radicle-surf/blob/fintan/add-diff-tests/src/diff/mod.rs#L438-L474

Basically it creates two directories. One is just the root and the `Repo` object (i.e. a fresh repository). The other is the root with the `Repo` and a new file in a directory `src/banana.rs`.

We should expect a `CreateFile` with a `Path` of `~/src/banana.rs` but it ends up only being `src/banana.rs`. 

My suspicion is that it is because we are not doing the algorithm relative to one of the directories. This hypothesis can be confirmed, at least in this case, by starting the `parent_path` off with `right.label`, since this passes the test. I'm going to explore further other test cases. Such as having a disjoint starting point, i.e. left starts with `bar` and right starts with `foo`.

FintanH:
> Such as having a disjoint starting point, i.e. `left` starts with `bar` and `right` starts with `foo`.

So more thoughts on this. If we have two directories `foo/src/banana.rs` and `bar/src/pineapple.rs` it's not easy to tell whether these are completely distinct or if we have done `mv banana.rs pineapple.rs` and `mv foo bar`. The only way this could definitely be known in our current model is if the contents of the two files are the exact same, and this needs to be propagated upwards to tell `foo -> bar`. 

At the moment it's impossible to tell whether if it the actions were `mv banana.rs pineapple.rs` followed by editing `pineapple.rs`.

atomashpolskiy:
> We should expect a CreateFile with a Path of ~/src/banana.rs but it ends up only being src/banana.rs.

I'd argue that we don't strictly need the paths to start with `~` (or the label of the enclosing directory in the general case). You might as well have passed two `~/src/` directories to `diff`. In such case you would end up with a `CreateFile {path: "src/mod.rs"}`, which is also not an root-based path.

Given that the (visual) renderer would probably prepend some prefix to the paths anyway (in order to show absolute file paths, e.g. `/Users/fintan/git-workspace/radicle-surf/src/banana.rs`), I'd say that relative paths make more sense. IMO, of course.

It might be a good moment to stop and think about this in terms of GUI. Let me illustrate my reasoning with a few screenshots from CLion (IDE that I use for Rust):

_Comparing entire repo between two revisions_
[How to get](https://ibb.co/Qp7jt5N)
[Result](https://ibb.co/2MRLFGs)

_Showing history of a single directory_
[How to get](https://ibb.co/Fmpwr0H)
[Result](https://ibb.co/hy1vGbX)

_Comparing two different directories in some revision_
[How to get](https://ibb.co/D5C6Vr6)
[Result](https://ibb.co/QmdT5LS)

Notice that all paths are relative, and this in particular makes it possible to compare identically named files in two _different_ directories.

atomashpolskiy:
> So more thoughts on this. If we have two directories foo/src/banana.rs and bar/src/pineapple.rs it's not easy to tell whether these are completely distinct or if we have done mv banana.rs pineapple.rs and mv foo bar. The only way this could definitely be known in our current model is if the contents of the two files are the exact same, and this needs to be propagated upwards to tell foo -> bar.

> At the moment it's impossible to tell whether if it the actions were mv banana.rs pineapple.rs followed by editing pineapple.rs.

I think this is true for all diff tools. I left a comment in the code on this issue. Basically, you need some heuristic to judge if the files _could have been moved_. My proposed solution is comparing pairs of files in DxC (cartesian product of Deleted and Created sets) looking for the longest common substring, but it can be more sophisticated (e.g. calculating the ratio of diff size to file size). We might look into existing tools for a clue.

FintanH:
> Notice that all paths are relative, and this in particular makes it possible to compare identically named files in two different directories.

Thanks for taking the time to illustrate what you mean! 

Looking through the examples and your last point there, I think this is what I'm trying to get at. You need paths to be relative from some point and in our case it would be the root, here symbolized as `~`, and in your GUI it's `/Users/etc.`

I'll give you a counter-example to show why we need to know some relative information. If we're given a tree like:
```
foo
|-- bar
|   `-- quux
|       `-- banana.rs
`-- baz
    `-- quux
        `-- banana.rs
```

And we pass in the two directories `left = quux/banana.rs`, `right = quux/banana.rs` how can we tell which parent it came from?

> I think this is true for all diff tools. I left a comment in the code on this issue. Basically, you need some heuristic to judge if the files could have been moved. My proposed solution is comparing pairs of files in DxC (cartesian product of Deleted and Created sets) looking for the longest common substring, but it can be more sophisticated (e.g. calculating the ratio of diff size to file size). We might look into existing tools for a clue.

Ah yes! Sorry you had left that comment :) Interesting thoughts :grin: Can definitely be left for another PR as well.

=========
Overall, I think this discussion is going well and we're asking some good questions of what we want! I think something that might help me understand your point of view is a more formal explanation of the features you want from the diff, followed by some examples. And obviously we can continue the conversation here too :)

I think something that can go in `docs` would be super helpful so that we can keep track of our decisions on this and have a resource for the decisions we made on this.

What do you think? :)

atomashpolskiy:
> And we pass in the two directories left = quux/banana.rs, right = quux/banana.rs how can we tell which parent it came from?

I don't quite understand why it might be important to know the parent in the first place. That's the conundrum, basically. As I see it, there's no use to this information; relative paths already allow us to make all kinds of comparisons.

And I also don't see the point of explicit `~`; the root is usually implicit, it does not have to have a name or be present in the path (unless it's required by path syntax, like the `/`; but in our case there's already a dedicated object for holding root contents, so you don't need to be able to address it with any symbolic name).

> Overall, I think this discussion is going well and we're asking some good questions of what we want! I think something that might help me understand your point of view is a more formal explanation of the features you want from the diff, followed by some examples. And obviously we can continue the conversation here too :)

> I think something that can go in docs would be super helpful so that we can keep track of our decisions on this and have a resource for the decisions we made on this.

> What do you think? :)

I think it's a good idea. I should note that I didn't intend to bloat the conversation, but rather thought that the examples of use cases (expressed in GUI of an existing tool) might clarify and kind of prove some of my points. Formalizing these use cases might definitely be beneficial, and in fact I think this can be done in source documentation (as a doc for the diff method or the diff module).

FintanH:
> I don't quite understand why it might be important to know the parent in the first place. That's the conundrum, basically. As I see it, there's no use to this information; relative paths already allow us to make all kinds of comparisons.

And I also don't see the point of explicit ~; the root is usually implicit, it does not have to have a name or be present in the path (unless it's required by path syntax, like the /; but in our case there's already a dedicated object for holding root contents, so you don't need to be able to address it with any symbolic name).

I could be the case that I'm just not seeing the point and need a little more guidance :) I'll do some more thinking around this tomorrow. Maybe come up with some more test cases and see what you think the outcomes should be.

> I think it's a good idea. I should note that I didn't intend to bloat the conversation, but rather thought that the examples of use cases (expressed in GUI of an existing tool) might clarify and kind of prove some of my points. Formalizing these use cases might definitely be beneficial, and in fact I think this can be done in source documentation (as a doc for the diff method or the diff module).

Awesome! And they did help a bit, and no worries about the conversation. I think this is super cool to get down to the details and tease out the implementation. Also, yes to inline docs! But I think I'd also like to something more "algorithm-y". For example, the whole model of the project was written in pseudo-Haskell [here](https://github.com/radicle-dev/radicle-surf/blob/master/docs/denotational-design.md). Something in the same vein would be good to have.

kim:
Maybe not the best timing for chiming in with some design considerations, but I’m wondering what our goals are for diffing.

We left the design at “tagged representation of the difference between two filesystem trees”. That’s fairly straightforward, except for the nuisance that git doesn’t record renames, and instead goes to great lengths inferring those after the fact. However, diffing two trees is only the “happy path” when both are a snapshot of the same linear history. More interesting — and arguably common — is to look at the diff between two branches: since they could have diverging histories, changes may be conflicting. Which means that their common ancestor needs to be taken into account.

Note that the semantics are subtly different for patch based systems: interestingly, you can’t take the diff between two branches in pijul, but only pick patches (in _some_ order) from one branch to be applied to another — which may also result in a conflict, but not in the same 3-way sense.

Note also that a conflict is not limited to edits _within_ a file, but also renames or deletes (consider a file was deleted on one side, but modified on the other).

So, while I think we should limit ourselves to diffs for now (as opposed to patches, which have the property of not producing conflicts within some context), we should probably flesh out the semantics of _what_ we’re taking the diff of some more. It’s worth pointing out that libgit provides a few different methods to obtain diffs, and that the representation is also tagged. Generalising this to also include pijul’s semantics is more challenging, although I can see that they distinguish between patch and diff, too.

FintanH:
> Maybe not the best timing for chiming in with some design considerations, but I’m wondering what our goals are for diffing.

Perfect timing really, because it's a good question :) And I think that's what I want to tease out as well.

> More interesting — and arguably common — is to look at the diff between two branches: since they could have diverging histories, changes may be conflicting. Which means that their common ancestor needs to be taken into account.

Are you implying here that we may need `History` information here for deriving the diff? 

> So, while I think we should limit ourselves to diffs for now (as opposed to patches, which have the property of not producing conflicts within some context), we should probably flesh out the semantics of what we’re taking the diff of some more. It’s worth pointing out that libgit provides a few different methods to obtain diffs, and that the representation is also tagged. Generalising this to also include pijul’s semantics is more challenging, although I can see that they distinguish between patch and diff, too.

Ya it might be worth exploring finding an interface for having different diff backends and you get the power from the backend itself. Maybe some more denotational design :eyes: 

===================

So, I think since this work is still valuable to even have as a reference, @atomashpolskiy: I can make a PR with some simple passing tests to your branch, you can tidy some things up if needs be and we can merge this PR.

Then the next steps are to have more of a discussion on what our options are for diffing and having some sort of formal semantics derived from that conversation. Does that sound reasonable to you @kim?

FintanH:
I've created an issue here https://github.com/radicle-dev/radicle-surf/issues/27

atomashpolskiy:
@FintanH , I think you can just rebase your changes against the PR branch, push the new commits to it directly (I checked "Allow edits from maintainers") and then merge.

FintanH:
@atomashpolskiy: Cool! I just wanted to give you an opportunity to flag anything that me seem weird about my PR and your implementation :)

FintanH:
> push the new commits to it directly

Either:
a) I somehow have forgotten my github password for authentication (but I don't think so)
b) I'm doing this wrong
c) Something else

But either way, I can't seem to push to your branch :sweat_smile: :

atomashpolskiy:
Then let's do as you've initially proposed :)

atomashpolskiy:
Wait, I think I can just merge your branch

atomashpolskiy:
Done

PR #30, opened 2019-12-10
Reexport git2 dependency

xla:
A lot of the git based implementations return git2 types, having the crate re-exported makes it safer for consumers to depend on them while depending on the same version.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,5 +60,3 @@ pub mod file_system;
 pub mod vcs;
 
 pub use crate::vcs::git;
-
-pub use git2;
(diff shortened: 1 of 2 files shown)


xla:
@FintanH AFAIK it has to end up on the top-level for external visibility.

FintanH:
@xla: Just did a test there, it works if you re-export from an internal module.

I think my preference would be `radicle_surf::vcs::git::git2`, so adding `pub use git2;` to that module.

xla:
@FintanH Great to know, thanks for the feedback. I think it should live down there as well, so it's nice and contained.

PR #32, opened 2019-12-10
Filter Branches

### LINKED ISSUE #29: When retrieving branches origin/HEAD is returned
What's the meaning of `origin/HEAD` when calling `list_branches`? In git `HEAD` is the symbolic reference which points to the tip of the current branch, or detached when checking out a tag. I don't think there is such a thing as a `HEAD` for a remote.

FintanH:
Fixes #29 

* Adds new `struct Branch`
  * A branch's name
  * It's `BranchType`
* `list_branches` now returns `Branch` so the caller can inspect `BranchType` if needs be
* `list_branches` can be filtered by `Local` or `Remote`

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -30,17 +30,15 @@
 //! ]);
 //! ```
 
+// Re-export git2 as sub-module
 pub use git2;
 
 use crate::file_system;
 use crate::vcs;
 use crate::vcs::VCS;
-use git2::{Commit, Error, Oid, Reference, Repository, TreeWalkMode, TreeWalkResult};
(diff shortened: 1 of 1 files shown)


src/vcs/git.rs
```diff
@@ -37,6 +37,9 @@ use git2::{Commit, Error, Oid, Reference, Repository, TreeWalkMode, TreeWalkResu
 use nonempty::NonEmpty;
 use std::collections::HashMap;
 
+// Re-export `BranchType` from here
+pub use git2::BranchType;
```
xla: Is that necessary after #30 landed?

FintanH: Ya not really :)

PR #34, opened 2019-12-10
Stable Testing

FintanH:
* Remove the test on empty tags
* Pin the lib example to a commit

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -72,16 +72,17 @@ impl<'repo> GitRepository {
     /// let repo = GitRepository::new("./data/git-golden").unwrap();
     /// let browser = GitBrowser::new(&repo).unwrap();
     ///
-    /// let branches = browser.list_branches();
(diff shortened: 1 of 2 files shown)


src/vcs/git.rs
```diff
@@ -72,16 +72,17 @@ impl<'repo> GitRepository {
     /// let repo = GitRepository::new("./data/git-golden").unwrap();
     /// let browser = GitBrowser::new(&repo).unwrap();
     ///
-    /// let branches = browser.list_branches();
+    /// let mut branches = browser.list_branches().unwrap();
+    /// branches.sort();
```
xla: :eyes: Sneaky!

FintanH: It ain't a `SortedVec` ;P

But if we do need some sort of ordering then we can create an issue. I think this is perfectly justified for testing purposes :)

xla: Absolutely.

PR #38, opened 2019-12-12
Passing root Path to last_commit produces None

xla:
This is a POC to show that when passing `Path::root()` to `last_commit` it results in `None`. It caused trouble for us when querying for the last commit of a tree that happen to be the repository root or in our terms an `""` prefix. I open this to pose the question if this is intended behaviour and if it's the callers responsibility to guard against that case.


FintanH:
Working test is here: https://github.com/radicle-dev/radicle-surf/blob/master/src/vcs/git.rs#L903-L920

PR #39, opened 2019-12-13
Checking is_root on a Path created with empty string is false

xla:
This is a POC with the help of a doc test showing when checking for `is_root` on a `Path::with_root(&["".into()])` is `false`. Is this intended behaviour? How should a caller should treat this case correctly?


FintanH:
I'd like to take a look at this again when #44 lands. Pretty sure it fixes this issue since we'll disallow empty labels, unless you use unsound and that would just be very un-sound of you.

xla:
@FintanH Sound plan - thought it might be addressed in #44

FintanH:
Use of `TryFrom` should prevent this due to #44

PR #40, opened 2019-12-13
Calling last_commit always shows latest instead of last one that touched the file

xla:
This POC shows how `last_commit` always returns the latest commit for the history instead of the last commit that touched that file. The expected behaviour is to return the commit which had a change for the path requested. See Github UI as a reference for different last commit in a listing of files.


FintanH:
Fixed in #52

PR #42, opened 2019-12-17
Relative Directory Ops

FintanH:
* Makes operation relative to the `Directory` the caller is in
* Should remove the use of `root` but will keep it in for backward-compat for now

### DIFF
--- a/src/file_system/mod.rs
+++ b/src/file_system/mod.rs
@@ -87,6 +87,10 @@ impl fmt::Display for Path {
 }
 
 impl Path {
+    pub fn new(label: Label) -> Path {
+        Path(NonEmpty::new(label))
+    }
+
     /// The root path is a `Path` made up of the single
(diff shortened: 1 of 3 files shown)


FintanH:
> Don’t have context - but shouldn’t it then be possible to also walk up the directory tree (a la “..”)?

It's mostly so that the user doesn't have to say the directory they're in, e.g. `~/foo/bar` when they're already in the root directory. Instead it would just be `foo/bar`. 

Not sure if we can go back up :thinking: You get a new directory back so in theory you still have the previous directory.

kim:
> You get a new directory back
>

Which you can pass around on it’s own. As I said I don’t have the context
on what led you to this, just sayin that this could be handy.

FintanH:
The context is ergonomics for our users :grin: And possibly not exposing the concept of the "root"

PR #44, opened 2020-01-02
Fintan/label conversion

### LINKED ISSUE #35: Use FromStr trait for Path
Can use [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) for _ergonomics_

FintanH:
Fixes #35 

Introduces parsing `str` to get `Label` and `Path`.

* Parsing `str` now return `Result`
* Introduces `unsound` module which is the escape hatch for making `Label`s and `Path`s without handling `Result`s all the time.
* Needed to add new logic in `git` to handle the `Result` from above.

### DIFF
--- a/src/file_system/error.rs
+++ b/src/file_system/error.rs
@@ -1,0 +1,35 @@
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Error {
+    Label(Label),
+    Path(Path),
+}
+
+impl From<Label> for Error {
+    fn from(err: Label) -> Self {
+        Error::Label(err)
+    }
+}
+
+impl From<Path> for Error {
+    fn from(err: Path) -> Self {
+        Error::Path(err)
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/file_system/error.rs
```diff
@@ -0,0 +1,35 @@
...
+impl From<Path> for Error {
+    fn from(err: Path) -> Self {
+        Error::Path(err)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Path {
+    Empty,
+    Label(Label),
```
xla: I understand that this enum is for future extensibility but it feels funky to have a `Label` variant in both `Error` and `Path` enums. It will lead to more complexity on the caller side.

FintanH: Good point :ok_hand: It did feel a bit janky writing it. Cleaned up now :+1:

PR #45, opened 2020-01-05
Implemented From<Vec<Label>> for Path.

massimiliano-mantione:


src/file_system/mod.rs
```diff
@@ -354,6 +354,23 @@ impl Path {
     }
 }
 
+impl From<Vec<Label>> for Path {
+    fn from(labels: Vec<Label>) -> Self {
+        if labels.len() > 0 {
+            Self::from_labels(labels[0].clone(), &labels[1..])
+        } else {
+            panic!("Cannot convert empty vector to Path")
```
xla: In my understanding best practice is to convert all fallible `From` implementations to `TryFrom`. There is even a lint in good ole clippy: https://rust-lang.github.io/rust-clippy/master/#fallible_impl_from

src/file_system/mod.rs
```diff
@@ -1152,4 +1169,43 @@ pub mod tests {
...
     }
+
+    fn path_from_string_vector(names: &Vec<&str>) -> Path {
+        let labels: Vec<Label> = names.iter().map(|n| Label::from(*n)).collect();
+        Path::from(labels)
+    }
+    
+    #[test]
+    fn build_path_from_vector() {
+        let names = vec!["a", "b", "c"];
```
xla: If I recall correctly other places in the repo prefer `bar`, `baz`, etc. for this kind of test enumeration.

src/file_system/mod.rs
FintanH: Ya, so I would always discourage code like this. `panic` should only be used for code that can truly exit, like a CLI or testing. 

@xla: Is a `TryFrom` instance something application would be interested in?

@massimiliano-mantione: The original issue for the repo was actually to go from `Path` to `Vec<Label>`. I probably worded it badly, so my bad. You'll notice that it's safe to have a function for `Path -> Vec<Label>` without any errors :)

src/file_system/mod.rs
```diff
@@ -354,6 +354,23 @@ impl Path {
...
+        } else {
+            panic!("Cannot convert empty vector to Path")
+        }
+    }
+}
+
+impl From<NonEmpty<Label>> for Path {
+    fn from(labels: NonEmpty<Label>) -> Self {
+        let (first, rest) = labels.split_first();
+        Self::from_labels(first.clone(), rest)
```
FintanH: This would actually be `Path(labels)`, since `Path` is just a newtype wrapper around `NonEmpty<Label>` :)

src/file_system/mod.rs
FintanH: This is true, I have been using those simplistic names, but don't really have guidelines :sweat_smile: Consistency makes sense though :ok_hand:

src/file_system/mod.rs
xla: > @xla: Is a `TryFrom` instance something application would be interested in?

Maybe, we almost always more likely to come around with a `String` so a `FromStr` sounds like a quality of life improvement.

FintanH: > Maybe, we almost always more likely to come around with a String so a FromStr sounds like a quality of life improvement.

Ya that makes sense, and that was added in #44

FintanH:
@massimiliano-mantione I'm going to close this. I added the `From<Path> for Vec<Label>` instance in another PR

PR #46, opened 2020-01-06
fix: last_commit for files

FintanH:
This is an initial fix for `last_commit`. It currently works on files but not on directories (since that logic is slightly more complicated).


FintanH:
@kim: I'm conscious that you might want to review this code, but if not lemme know and I'll merge :)

PR #51, opened 2020-01-14
Improve name consistency for public git entities

xla:
With the drop of the `Git` prefix we move towards imports which should be qualified. As `git` is already in the module the prefix should be dropped consistently. Additionally `git2` is reexported again.

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -37,7 +37,7 @@
 //! ```
 
 // Re-export git2 as sub-module
-pub use git2::{BranchType, Error as GitError, Oid, Time};
+pub use git2;
 
 pub mod error;
 
@@ -45,7 +45,6 @@ use crate::file_system;
 use crate::vcs;
 use crate::vcs::git::error::*;
 use crate::vcs::VCS;
-use git2;
 use nonempty::NonEmpty;
 use std::cmp::Ordering;
 use std::collections::HashMap;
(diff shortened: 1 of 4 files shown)


src/vcs/git.rs
```diff
@@ -37,15 +37,14 @@
 //! ```
 
 // Re-export git2 as sub-module
-pub use git2::{BranchType, Error as GitError, Oid, Time};
```
FintanH: I would still like to keep these actually. Make it easier for the user since these are directly used in the surf API

xla:
> Suggestion: use qualified import in docs to show that that's what we expect, but not necessary if it's too much of a PITA for now.

I tried where applicable, for the doc(tests) in the `git` module it seems fine to be unqualified.

src/vcs/git.rs
xla: See 763eace.

FintanH:
> I tried where applicable, for the doc(tests) in the git module it seems fine to be unqualified.

Ya, it compiles but I mean that docs are how we talk to users of the library so the import should be qualified. But I can do it at a later date :)

xla:
> Ya, it compiles but I mean that docs are how we talk to users of the library so the import should be qualified. But I can do it at a later date :)

That was clear to me from the getgo, my point is that for the docs in the context of the `git` module it makes sense to have them unqalified, I'd advocate for the qualified in README and crate docs.

FintanH:
> That was clear to me from the getgo, my point is that for the docs in the context of the git module it makes sense to have them unqalified, I'd advocate for the qualified in README and crate docs.

Ah, I see! I think it would be good to be consistent across the docs, as the user will click around. But no worries for now. I'll merge and make a release.

PR #52, opened 2020-01-14
Last Commit

FintanH:
I wasn't happy with the interface for the tree so I changed it:
* `Node` now also has a `key: K`
* This changed the business logic a bit but the semantics stayed the same
* Added `insert_with`

Now `last_commit` works on directories and files :tada:

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -207,7 +207,7 @@ impl<'repo> Repository {
             .ok_or(Error::EmptyCommitHistory)
     }
 
-    fn last_commit(
+    fn file_history(
         &'repo self,
         commit: Commit,
     ) -> Result<Forest<file_system::Label, NonEmpty<OrderedCommit>>, Error> {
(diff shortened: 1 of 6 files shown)


src/vcs/git.rs
```diff
@@ -852,17 +887,38 @@ impl Browser {
...
+    /// use radicle_surf::file_system::unsound;
+    ///
+    /// let repo = Repository::new("./data/git-platinum")
+    ///     .expect("Could not retrieve ./data/git-platinum as git repository");
+    /// let mut browser = Browser::new(repo).expect("Could not initialise Browser");
+    ///
+    /// let expected_commit_id =
+    ///     Oid::from_str("3873745c8f6ffb45c990eb23b491d4b4b6182f95").unwrap();
+    ///
+    /// let gitignore_last_commit_id = browser
```
xla:
```suggestion
    /// let root_last_commit_id = browser
```

src/vcs/git.rs
```diff
@@ -852,17 +887,38 @@ impl Browser {
...
+    ///     .last_commit(&Path::root())
+    ///     .expect("Failed to get last commit")
+    ///     .map(|commit| commit.id);
+    ///
+    /// assert_eq!(gitignore_last_commit_id, Some(expected_commit_id));
```
xla:
```suggestion
    /// assert_eq!(root_last_commit_id, Some(expected_commit_id));
```

PR #58, opened 2020-01-17
Directory Tree

### LINKED ISSUE #25: Store DirectoryContents in sorted order

FintanH:
This is a _bit_ of a reorg of surf :grimacing: 

I've modularised things within `file_system`
  * `path`
  * `path::unsound`
  * `directory`

`Directory` is now backed by `Forest` from `tree.rs`. Actually makes for creating testing a lot easier and the interface seems nicer :+1: 

Fixes #25 since the tree is kept in sorted order


src/diff/mod.rs
```diff
@@ -102,65 +97,81 @@ impl Diff {
...
-                                (Repo, Repo) => {
-                                    old_entry_opt = old_iter.next();
-                                    new_entry_opt = new_iter.next();
-                                }
-                                (Repo, _) => {
-                                    old_entry_opt = old_iter.next();
-                                }
-                                (_, Repo) => {
-                                    new_entry_opt = new_iter.next();
+                        Ordering::Equal => match (new_entry, old_entry) {
```
xla: These triple-nested match shenanigans become quite hard to dissect as a reader, maybe the big body can be broken apart.

FintanH:
> When is the turning point for usefulness of doctests, while in prinicple I'm really into the idea, some of these blocks are massive and I wonder how easy it will be to maintain going forward? Or if some of these are better surfed with normal test cases.

On the one hand, I think the doc tests are super useful and I appreciate how someone can copy and paste code and then rework it for their use case.

On the other hand, I agree that they can become unwieldy and annoying to scroll through. I think something to look into is maybe having shared blocks where we can define a block of code for setting up a test (e.g. Creating a shared directory structure for `find_file`, `find_directory`, `insert_file`, etc.). I think this might be possible but not 100% sure. I'll create a ticket for this.

PR #61, opened 2020-01-22
Revparse Refactor

### LINKED ISSUE #55: Smart Revisions
Currently it's a bit verbose for users to select a generic revision: https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/schema/mod.rs#L174-L177

Let's get smart.

FintanH:
Fixes #55 

Introduce `RevObject` which is created via https://git-scm.com/docs/git-rev-parse.html#_specifying_revisions.

This surfaced some silly things in the API so I removed some things that aren't needed such as Object, Sha1, and de-duplicated some naming.

All the object related code is now under `vcs::git::object`.

Most importantly @xla: you can replace [this](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/schema/mod.rs#L111-L115) with `browser.revspec(revision)` :)

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -7,6 +7,7 @@ steps:
   command:
   # Ensure we have fetched the remote info in submodule
   - git submodule foreach "git fetch origin"
+  - git submodule update --remote
   - "cargo test --all"
   agents:
     production: "true"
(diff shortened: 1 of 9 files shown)


FintanH:
Since `radicle-link` has the commit sha in its master and no one is reviewing I'm going to merge :upside_down_face:

Issue #77, opened 2020-03-13
Error Tests


FintanH:
[This](https://rust-lang.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark) is soooo much better than using `expect` and `unwrap` in tests. Let's clean them up making our docs a lot better!

sonro:
Hi there, I would like to help with this. I see you don't have a contributing guide or a licence, are there any speical requirements you have before I begin a pull request?

FintanH:
Hey @sonro :wave: Really glad you'd like to help :)

Contribution guidelines are not up yet, but we have a ticket for it: https://github.com/radicle-dev/radicle-surf/issues/33. If you notice anything that's difficult to get set up or run it would be awesome to have it tracked in that issue! Sometimes it's hard to see the wood from the trees.

Regarding the license. We have marked it as GPL-3.0-only in the [Cargo.toml](https://github.com/radicle-dev/radicle-surf/blob/master/Cargo.toml#L10). There are a few more steps to take which I just added a ticket for here: https://github.com/radicle-dev/radicle-surf/issues/85.

Thanks for bringing these up, and let me know if I can do anything to help :v:

Issue #76, opened 2020-03-13
Error Libraries and Practices


FintanH:
Following [these conventions](https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err) as well as our own radicle-dev/radicle-decisions#2.

We should use `thiserror` and make some errors more opaque and consistent.

FintanH:
The last thing to do is add `non_exhaustive`

sonro:
I can update this.

FintanH:
Thanks! I should have remembered when I was reviewing :sweat_smile:

PR #87, opened 2020-03-16
Add 'thiserror' library

sonro:
In order to update the doctests to be in line with the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark) (see #77), this crate's error types need to be updated.

This pull request converts existing error types to conform with the [thiserror](https://crates.io/crates/thiserror) API: see #76. There are no further adjustments to the structure of the errors, however now that they all implement `std::error::Error` and `std::fmt::Display`.

Once this change is merged I can open an additional pull request to fix the doctests.

### DIFF
--- a/src/file_system/error.rs
+++ b/src/file_system/error.rs
@@ -3,6 +3,8 @@
 //! These errors occur due to [`Label`] and [`Path`] parsing when using their respective `TryFrom`
 //! instances.
 
+use thiserror::Error;
+
 pub(crate) const EMPTY_PATH: Error = Error::Path(Path::Empty);
 
 pub(crate) const INVALID_UTF8: Error = Error::Label(Label::InvalidUTF8);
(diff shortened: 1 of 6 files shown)


src/file_system/error.rs
```diff
@@ -3,47 +3,43 @@
...
     Empty,
 }
 
 /// Parse errors for when parsing a string to a [`Label`].
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
 pub enum Label {
     /// An error signifying that a [`Label`] is contains invalid UTF-8.
+    #[error("Label contains invalid UTF-8")]
     InvalidUTF8,
```
FintanH: Something I think we should do here is to actually have a `Label` inside `InvalidUTF8` an `ContainsSlash` so that we can provide more detail as to which `Label` was wrong making our errors more informative :) Could you add this while we're here editing this code?

FintanH:
I'm currently doing some fixing of our buildkite builds. I'll let you know when that's working :)

src/file_system/error.rs
sonro: Great idea, editing this now.

src/file_system/path/mod.rs
```diff
@@ -374,7 +374,7 @@ impl TryFrom<path::PathBuf> for Path {
     fn try_from(path_buf: path::PathBuf) -> Result<Self, Self::Error> {
         let mut path = Path::root();
         for p in path_buf.iter() {
-            let p = p.to_str().ok_or(error::INVALID_UTF8)?;
+            let p = p.to_str().ok_or(error::label_invalid_utf8(p))?;
```
FintanH:
```suggestion
            let p = p.to_str().ok_or(|| error::label_invalid_utf8(p))?;
```

FintanH:
Thanks again @sonro!

PR #88, opened 2020-03-18
Add Licensing + Format Cleanup

FintanH:
Rearrange CI scripts for a cleaner organisation.
We are also adding the necessary licensing and `rustfmt` checks to ensure
that the licensing information is present.

Since we are editing the files to add the license headers we're also
moving the mod.rs files to their named equivalents to be in line with
Rust 2018 conventions.


deny.toml
```diff
@@ -0,0 +1,214 @@
+# This section is considered when running `cargo deny check advisories`
+# More documentation for the advisories section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html
+[advisories]
+# The path where the advisory database is cloned/fetched into
+db-path = "/cache/cargo/advisory-db"
```
kim: I added this: https://github.com/radicle-dev/radicle-link/pull/75/files#diff-8d3288bb6c44ba57934db1e4fa9c1d97R10-R13

So one can run `cargo deny` locally without any surprises

scripts/license-headers
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+#
+# Iterates over all rust source files and ensures they start with the license
+# header as per the `.license-header` file at the root of the repository.
+
+set -euo pipefail
+IFS=$'\n'
+
+shopt -s globstar
+for file in */src/**/*.rs
```
kim: I just realised that this will miss out on tests and examples, so changed it to this:

```suggestion
shopt -s globstar extglob
for file in */+(src|tests|examples)/**/*.rs
```

FintanH:
I borked this PR... Going to open a new one

PR #90, opened 2020-03-26
License Headers

FintanH:
Take 2


FintanH:
For some reason, I can't shift the build off of `mac-n-cheese` :thinking: I tried setting the agent name in the "New Build" console :man_shrugging:

kim:
> `mac-n-cheese`

I'm guessing that the `pipeline upload` step was scheduled there. This one can only be configured in the BK UI, and there you need to set "Agent Targeting Rules" to something which won't match the mac. I did that already.

PR #97, opened 2020-05-12
Bump git2 crate to latest version (0.13.x)

kim:
Also ensure openssl is not pulled in, as the openssl license is not compatible with the GPL. Note that the rust bindings crates are themselves MIT/Apache-2 licensed, but that does not seem to be legit, as they link against the native openssl library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,9 +10,13 @@ repository = "https://github.com/radicle-dev/radicle-surf"
 license = "GPL-3.0-or-later"
 
 [dependencies]
-git2 = "0.12"
 thiserror = "1.0"
 nonempty = "0.4"
+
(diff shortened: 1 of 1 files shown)


kim:
This is getting a litte bit annoying -- either `git2` is not following SemVer, or they are moving really fast. Either way, I would at this point prefer to either set the version bound to `^0` (smaller than `1.0`), or move `radicle-surf` into the `radicle-link` repo.

FintanH:
It seems they jumped from `12` to `13`. I'd be happy with `^0`. I'd also be happy with moving it to `radicle-link` but it'd be a major pain to move the issues :/

FintanH:
I'm just gonna merge this. I can switch the versioning in the next PR

PR #102, opened 2020-05-14
Git diff

cloudhead:


src/vcs/git.rs
```diff
@@ -192,6 +205,77 @@ impl<'repo> Repository {
...
+                    diff.add_created_file(&label, &parent);
+                },
+                Delta::Deleted => {
+                    let diff_file = delta.old_file();
+                    let path = diff_file.path().expect("file should have a path");
+                    let name = path.file_name().expect("path should have a name");
+                    let parent = path.parent().expect("path should have a parent");
+
+                    let label = file_system::Label::try_from(&*name.to_string_lossy())?;
+                    let parent = file_system::Path::try_from(parent.to_path_buf())?;
```
cloudhead, line 235: This dance should be factored out somewhere, and maybe it can be re-written differently :roll_eyes:

src/diff.rs
```diff
@@ -61,11 +61,75 @@ pub struct ModifiedFile {
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct Hunk {
+    pub header: Line,
+    pub lines: Vec<LineDiff>,
+}
+
+/// The content of a single line.
+pub type Line = Vec<u8>;
```
cloudhead, line 74: I wonder if we should use a ref instead of an owned type here. Less copying for large diffs, but maybe more of a pain to use. Though this is how `git2` works.

src/vcs/git.rs
cloudhead: I tried `file_system::Path::try_from()`, but it's also pretty painful.

src/diff.rs
```diff
@@ -61,11 +61,75 @@ pub struct ModifiedFile {
...
+/// addition and one deletion. Context is also represented with this type.
+#[derive(Debug, PartialEq, Eq)]
+pub struct LineDiff {
+    /// Line number in old file or `None` for added line.
+    pub old_line_num: Option<u32>,
+    /// Line number in new file or `None` for deleted line.
+    pub new_line_num: Option<u32>,
+    /// Line content.
+    pub line: Line,
+    /// Sigil showing origin of this `LineDiff`. Eg. `+`, `-`, etc.
```
FintanH, line 86: What if we made an enum here instead:
```rust
enum LineDiff {
  Addition { line_num: u32, line: Line }, // this is the new line
  Removal { line_num: u32, line: Line }, // this was the old line
  Modification { line_num: u32, old: Line, new: Line },
}
```

This way we're not matching on a `char` :grin: (well apart from when we parse)

FintanH: Oh and `Context` too.

src/diff.rs
```diff
@@ -61,11 +61,75 @@ pub struct ModifiedFile {
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct FileDiff {
-    // TODO
+    pub hunks: Vec<Hunk>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct Hunk {
```
FintanH, line 68: I'd be interested in thinking of a less `git`ty name. But not opposed to keeping it as `Hunk` either. My suggestion would something along the lines of `FileChange`

src/diff.rs
FintanH: The proxy would have to do a clone anyway to send it to the front end right?

src/diff.rs
cloudhead: So I thought about this, but we can actually match on the old/new line:

`(Some(old), None)` is a deletion
`(None, Some(new))` is an addition
`(Some(old), Some(new))` is context

Putting the whole thing in an enum makes it a bit annoying to retrieve the lines. Also keep in mind that there is no `Modification`, because those would be represented as *two* `LineDiff`s.

src/diff.rs
cloudhead: I'm down, but not sure about introducing "Change" which is kind of another word for "Diff" :thinking: - ie. it'll sound like `FileDiff` which we already have and `FileChange` are the same thing.

src/diff.rs
cloudhead: So we'd end up with:

```rust
enum LineDiff {
  Addition { line_num: u32, line: Line }, // this is the new line
  Removal { line_num: u32, line: Line }, // this was the old line
  Context { line_num: u32, line: Line },
}
```

which has the benefit of being able to do:
```
match line_diff {
  Removal { line_num } | Context { line_num } | Addition { line_num } => ...
}
```

But it could just as well be represented as:
```
struct LineDiff {
  line: Line,
  kind: LineDiffKind
}
enum LineDiffKind {
  Addition, Deletion, Context
}
```
:thinking:

FintanH: I feel like remembering the order of `Some` and `None` in the tuple is a bit of a burden on the user :sweat_smile: And good point about `Modification` :)

I usually opt for the former because it's in canonical form -- sum of products. As opposed to the latter which is product of sums. But I don't think matters much.

My main goal here is to show the intent of the data through the types, so as long as we're describing the cases in a closed way I'll be happy and so will our consumers :)

src/diff.rs
FintanH: Maybe something like `DiffSet`?

src/vcs/git.rs
FintanH: What's the pain point here? I'd like to improve it if possible :) `Path` is there to ensure we have `NonEmpty`-ness and making sure user input is sanitised somewhat.

src/diff.rs
cloudhead: Yeah or perhaps `LineDiffSet`

src/vcs/git.rs
```diff
@@ -118,6 +119,17 @@ impl From<OrderedCommit> for Commit {
...
+    fn from(line: git2::DiffLine) -> Self {
+        match (line.old_lineno(), line.new_lineno()) {
+            (None, Some(n)) => Self::addition(line.content().to_owned(), n),
+            (Some(n), None) => Self::deletion(line.content().to_owned(), n),
+            (Some(_), Some(n)) => Self::context(line.content().to_owned(), n),
+            (None, None) => unreachable!(),
+        }
+    }
+}
+
```
cloudhead, line 132: `TryFrom`

src/diff.rs
cloudhead: Yeah true

src/vcs/git.rs
```diff
@@ -192,6 +204,65 @@ impl<'repo> Repository {
...
+                            for l in 0..hunk_lines {
+                                let line = patch.line_in_hunk(h, l)?;
+                                lines.push(line.into());
+                            }
+                            hunks.push(Hunk { header, lines });
+                        }
+                        diff.add_modified_file(path, hunks);
+                    } else if diff_file.is_binary() {
+                        // TODO
```
cloudhead, line 253: How should we handle this?

src/vcs/git.rs
```diff
@@ -192,6 +204,65 @@ impl<'repo> Repository {
...
+                    } else {
+                        unreachable!()
+                    }
+                },
+                _ => {},
+            }
+            //
```
cloudhead, line 260: ?

src/vcs/git.rs
```diff
@@ -192,6 +204,65 @@ impl<'repo> Repository {
...
+
+        let mut diff = Diff::new();
+        let git_diff = self.diff_commits(from, Some(to))?;
+
+        for (idx, delta) in git_diff.deltas().enumerate() {
+            match delta.status() {
+                Delta::Added => {
+                    let diff_file = delta.new_file();
+                    let path = diff_file.path().expect("file should have a path");
+                    let path = file_system::Path::try_from(path.to_path_buf()).unwrap();
```
cloudhead, line 219: Change to `expect()`?

src/diff.rs
```diff
@@ -61,11 +67,75 @@ pub struct ModifiedFile {
...
```
cloudhead, line 74: Document this

src/vcs/git.rs
FintanH: I think we can `?`. I'm pretty sure we have file path errors in the top-level `Error` type. I'd like to remove `expect` above as well :)

src/vcs/git.rs
FintanH: Should we have a `Diff::Binary` and not present those diffs? Just have two blobs?

src/vcs/git.rs
FintanH: These ones seem reasonable to handle:
```
    Modified,
    Renamed,
    Copied,
```

For the others, it might be an idea to introduce a new error variant `DeltUnhandled(git2::Delta)`, just so we can jeep track of one of these popping up.

src/vcs/git.rs
cloudhead: Yeah, changed `FileDiff` to an enum. However for now I'm not including the blobs in the Binary, since they'd have to be owned. I'm a little worried that this runs on a large file and all that data has to be transfered even though we aren't displaying it. I think we should handle that case separately and think more about the use-case first.

cloudhead: c5cc0a2

src/vcs/git.rs
cloudhead: c5cc0a2

src/diff.rs
```diff
@@ -53,23 +60,97 @@ pub struct MoveFile {
...
+    /// Line number in old or new file.
+    pub line_num: u32,
+    /// Line content.
+    pub line: Line,
+    /// Line diff kind, eg. addition or deletion.
+    pub kind: LineDiffKind,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum LineDiffKind {
```
FintanH, line 105: Could you document these? I'm adding a ci check on another branch to run `cargo doc --no-deps` so it'd be good to anticipate that :)

src/diff.rs
```diff
@@ -207,12 +290,15 @@ impl Diff {
...
+        F: Fn(Path) -> T + Copy,
     {
         match entry {
             DirectoryContents::Directory(dir) => Diff::collect_files(dir, parent_path, mapper),
             DirectoryContents::File { name, .. } => {
-                let mapped = mapper(name, &RefCell::borrow(parent_path));
+                let mut path = parent_path.borrow().clone();
+                path.push(name.clone());
+
+                let mapped = mapper(path);
```
FintanH, line 301: Let's clean this up while we're here and just have:
```rust
Ok(vec![mapper(path)]);
```

src/vcs/git.rs
```diff
@@ -118,6 +119,19 @@ impl From<OrderedCommit> for Commit {
...
 
+impl<'a> TryFrom<git2::DiffLine<'a>> for LineDiff {
+    type Error = &'static str;
+
+    fn try_from(line: git2::DiffLine) -> Result<Self, Self::Error> {
+        match (line.old_lineno(), line.new_lineno()) {
+            (None, Some(n)) => Ok(Self::addition(line.content().to_owned(), n)),
+            (Some(n), None) => Ok(Self::deletion(line.content().to_owned(), n)),
+            (Some(_), Some(n)) => Ok(Self::context(line.content().to_owned(), n)),
+            (None, None) => Err("invalid `git2::DiffLine`"),
```
FintanH, line 130: I'd like if we created a `non_exhaustive` error enum for `Diff`s :)

src/vcs/git.rs
```diff
@@ -192,6 +206,103 @@ impl<'repo> Repository {
...
+        let mut diff = Diff::new();
+        let git_diff = self.diff_commits(from, Some(to))?;
+
+        for (idx, delta) in git_diff.deltas().enumerate() {
+            match delta.status() {
+                Delta::Added => {
+                    let diff_file = delta.new_file();
+                    let path = diff_file
+                        .path()
+                        .ok_or(Error::GitDiff("couldn't retrieve file path"))?;
```
FintanH, line 222: If we did the above enum this would get embedded in the top-level `Error`

FintanH: Also, I think having the file path in the payload is useful for the error message. I don't want to be asking myself "which file path" in a few months time :smile:

src/vcs/git.rs
```diff
@@ -192,6 +206,103 @@ impl<'repo> Repository {
...
+                    let path = file_system::Path::try_from(path.to_path_buf())?;
+
+                    let patch = Patch::from_diff(&git_diff, idx)?;
+
+                    if let Some(patch) = patch {
+                        let mut hunks: Vec<Hunk> = Vec::new();
+
+                        for h in 0..patch.num_hunks() {
+                            let (hunk, hunk_lines) = patch.hunk(h)?;
+                            let header = hunk.header().to_owned();
```
FintanH, line 250: I was thinking, maybe we should have a type fo Header as well. What do you think? If we do want one then we can put it as a follow-up ticket -- doesn't need to be done now.

src/vcs/git.rs
cloudhead: That's a bit of a conundrum, since the error is due to the path not being available!

FintanH: Ah hahaha silly me! Disregard my madness

src/vcs/git.rs
cloudhead: Would just be an alias right? Not sure what these look like in pijul etc.

FintanH: That's what I'm unsure about. I'll create a follow-up ticket to figure exactly this out :)

PR #105, opened 2020-05-15
Add branch_contains function to Browser

### LINKED ISSUE #94: Branch information inside `Commit` object
For the application, we are going to need to know which branch a given commit is in. Interestingly, a commit can be in multiple branches. If this is the case, we need to pick one branch, but I'm not sure what kind of heuristic we can use here, except checking if the commit is part of the default branch. Looking at the GitHub UI, it only ever shows one branch per commit.

Perhaps this isn't surf's concern and it can at least for now just return the output of `git branch --contains <commit>`.

FintanH:
Fixes #94

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -362,6 +362,39 @@ impl<'repo> Repository {
         }
     }
 
+    pub(crate) fn branch_contains(&self, commit: &Oid) -> Result<Vec<Branch>, Error> {
+        let branches = self
+            .0
+            .branches(Some(BranchType::Local))?
+            .collect::<Result<Vec<(git2::Branch, BranchType)>, git2::Error>>()?;
+
+        let mut contained_branches = vec![];
+
+        branches.into_iter().try_for_each(|(branch, locality)| {
+            self.reachable_from(&branch.get(), &commit)
+                .and_then(|contains| {
+                    if contains {
+                        let branch = Branch::from_git_branch(branch, locality)?;
+                        contained_branches.push(branch);
+                    }
+                    Ok(())
+                })
+        })?;
+
+        Ok(contained_branches)
+    }
+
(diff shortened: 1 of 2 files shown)


src/vcs/git.rs
```diff
@@ -269,6 +269,39 @@ impl<'repo> Repository {
         }
     }
 
+    pub(crate) fn branch_contains(&self, commit: &Oid) -> Result<Vec<Branch>, Error> {
```
cloudhead, line 272: I expected this function to be a `(branch, rev) -> bool` function, given its name, ie. tell me if this branch contains this rev. I think we should try to find a better name for it, or possibly move it to the `Commit`? Brainstorming:
```
commit.branches() -> Vec<Branch>
repo.branches_with(rev) -> Vec<Branch>
repo.commit_branches(commit) -> Vec<Branch>
```

FintanH: It can't really go on a `Commit` because a `Commit` doesn't know about search history.

Changing it to accept revisions could definitely, given it resolves to a `Commit` of course :thinking:

cloudhead: yeah then I guess `Repository::commit_branches` or `revision_branches` might be clearer?

FintanH: Yup makes sense!

ci/run
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -eou pipefail
 
+# Want to ensure that dev and master have been
+# checked out for testing purposes
+echo '--- Checkout branches'
+git submodule foreach "git checkout dev"
+git submodule foreach "git checkout master"
```
FintanH, line 8: note to self: This is a better way of doing the checkout https://github.com/radicle-dev/radicle-surf/pull/106/files#diff-a7b0a2dee0126cddf994326e705a91eaR24

PR #106, opened 2020-05-18
RepositoryRef

FintanH:
The repo module is just moving the Repository code from git and then reorganising the functionality so that it works on RepositoryRef instead.

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -65,484 +65,30 @@
 // Re-export git2 as sub-module
 pub use git2::{self, BranchType, Error as Git2Error, Oid, Time};
 
+mod repo;
+pub use repo::{History, Repository, RepositoryRef};
+
 pub mod error;
 mod object;
 
-pub use crate::vcs::git::object::*;
+pub use crate::{
+    diff::Diff,
+    vcs::git::object::{Branch, BranchName, Commit, RevObject, Signature, TagName},
+};
+
 use crate::{
-    diff,
-    diff::*,
     file_system,
     file_system::directory,
-    tree::*,
     vcs,
     vcs::{git::error::*, VCS},
 };
 use nonempty::NonEmpty;
-use std::{cmp::Ordering, collections::HashMap, convert::TryFrom, str};
-
-/// A `History` that uses `git2::Commit` as the underlying artifact.
-pub type History = vcs::History<Commit>;
-
-/// Wrapper around the `git2`'s `git2::Repository` type.
-/// This is to to limit the functionality that we can do
-/// on the underlying object.
-pub struct Repository(pub(crate) git2::Repository);
-
-/// OrderedCommit is to allow for us to identify an ordering of commit history
-/// as we enumerate over a revwalk of commits, by assigning each commit an
-/// identifier.
-#[derive(Clone)]
-struct OrderedCommit {
-    id: usize,
-    commit: Commit,
-}
-
-impl std::fmt::Debug for OrderedCommit {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
-            f,
-            "OrderedCommit {{ id: {}, commit: {} }}",
-            self.id, self.commit.id
-        )
-    }
-}
-
-impl OrderedCommit {
-    fn compare_by_id(&self, other: &Self) -> Ordering {
-        self.id.cmp(&other.id).reverse()
-    }
-}
-
-impl From<OrderedCommit> for Commit {
-    fn from(ordered_commit: OrderedCommit) -> Self {
-        ordered_commit.commit
-    }
-}
-
-impl<'repo> Repository {
-    /// Open a git repository given its URI.
-    ///
-    /// # Errors
-    ///
-    /// * [`error::Error::Git`]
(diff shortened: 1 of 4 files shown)


src/vcs/git.rs
```diff
@@ -65,362 +65,26 @@
...
-        write!(f, ".git")
-    }
-}
+use std::{collections::HashMap, convert::TryFrom, str};
 
 /// A [`crate::vcs::Browser`] that uses [`Repository`] as the underlying
 /// repository backend, [`git2::Commit`] as the artifact, and [`Error`] for
 /// error reporting.
-pub type Browser<'a> = vcs::Browser<'a, Repository, Commit, Error>;
+pub type Browser<'a> = vcs::Browser<RepositoryRef<'a>, Commit, Error>;
```
kim, line 87: :+1:

src/vcs/git.rs
```diff
@@ -445,9 +109,10 @@ impl<'a> Browser<'a> {
     /// # }
     /// ```
     pub fn new(repository: &'a Repository) -> Result<Self, Error> {
+        let repository = repository.as_ref();
```
kim, line 112: We also need a constructor which takes `RepositoryRef`.

src/vcs/git/repo.rs
```diff
@@ -0,0 +1,388 @@
...
+
+/// A `History` that uses `git2::Commit` as the underlying artifact.
+pub type History = vcs::History<Commit>;
+
+/// Wrapper around the `git2`'s `git2::Repository` type.
+/// This is to to limit the functionality that we can do
+/// on the underlying object.
+pub struct Repository(pub(super) git2::Repository);
+
+/// A reference-only `Repository`. This means that we cannot mutate the
```
kim, line 68: The description doesn't quite match my understanding: `Send` only means that the value can be transferred across thread boundaries, _not_ that it can be shared among threads.

`git2::Repository` is `Send`, but to make it `Sync` (ie. shareable), we need to slap an `Arc<Mutex<git2::Repository>>`. If we wanted to get a `Browser` while holding the lock, we _should_ be able to something like this:

```rust
let locked_repo: MutexGuard<git2::Repository> = shared_repo.lock().unwrap();
let surf_repo = RepositoryRef(&locked_repo);
let bro = Browser::new(surf_repo);
// can do bro things while `locked_repo` is in scope
```

src/vcs/git/repo.rs
```diff
@@ -0,0 +1,388 @@
...
+/// underlying `Repository`. Not being able to mutate the `Repository` means
+/// that the functions defined for `RepositoryRef` should be thread-safe.
+pub struct RepositoryRef<'a> {
+    pub(super) repo_ref: &'a git2::Repository,
+}
+
+// RepositoryRef should be safe to share among threads since it is read-only
+unsafe impl<'a> Send for RepositoryRef<'a> {}
```
kim, line 76: Hm, `git2::Repository` is `Send`, would `&git2::Repository` not be?

src/vcs/git.rs
FintanH: I think I'll make `new` take `RepositoryRef` and change the call-sites.

FintanH: Or... Implement `AsRef` :thinking:

src/vcs/git/repo.rs
FintanH: Right, sorry. I should have read the docs first :grimacing:

src/vcs/git/repo.rs
FintanH: It wasn't being implemented by the compiler, so I slapped this down here.

src/vcs/git.rs
FintanH: Used the `Into` trick instead :ok_hand:

src/vcs/git/repo.rs
kim: The question is if this really works. Perhaps we should add a test somewhere.

FintanH: Test sounds good :sweat_smile:

FintanH: https://github.com/radicle-dev/radicle-surf/pull/106/commits/ffa8f0747ef64f7c4d4fd7f75fefde906d141f4c

src/vcs/git/repo.rs
```diff
@@ -0,0 +1,521 @@
...
+pub struct Repository(pub(super) git2::Repository);
+
+/// A reference-only `Repository`. This means that we cannot mutate the
+/// underlying `Repository`. Not being able to mutate the `Repository` means
+/// that the functions defined for `RepositoryRef` should be thread-safe.
+pub struct RepositoryRef<'a> {
+    pub(super) repo_ref: &'a git2::Repository,
+}
+
+impl<'a> Deref for RepositoryRef<'a> {
```
kim, line 77: Ah wait, I think this is a bit dangerous as it could break the readonly contract

FintanH: But it returns a reference so it's still read-only no?

FintanH: Ah no, you're right. There's `&self` functions that mutate the underlying repository data

FintanH: Dropped the commit

PR #109, opened 2020-05-19
File/Directory History

### LINKED ISSUE #64: Commit History
View the commit history of a file or directory.

The majority of this work is done via `last_commit`, we just need to make the functionality `pub`.

FintanH:
Fixes #64 

Function for fetching the entire history for a file or directory.

@cloudhead: I'd like to get this in before I start optimising the history related bottleneck, i.e. last commit

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -560,16 +560,12 @@ impl<'a> Browser<'a> {
     /// # }
     /// ```
     pub fn last_commit(&self, path: file_system::Path) -> Result<Option<Commit>, Error> {
-        let file_history = self.repository.file_history(self.get().first().clone())?;
-
-        Ok(file_history.find(path.0).map(|tree| {
-            tree.maximum_by(&|c: &NonEmpty<repo::OrderedCommit>, d| {
-                c.first().compare_by_id(&d.first())
-            })
-            .first()
-            .commit
-            .clone()
-        }))
+        let file_history = self.repository.file_history(
+            &path,
+            repo::CommitHistory::Last,
+            self.get().first().clone(),
+        )?;
+        Ok(file_history.first().cloned())
     }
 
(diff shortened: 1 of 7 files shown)


src/vcs/git.rs
```diff
@@ -572,6 +572,85 @@ impl<'a> Browser<'a> {
...
+        self.repository
+            .file_history(self.get().first().clone())
+            .and_then(|history| {
+                let subtree = history
+                    .find(path.clone().0)
+                    .ok_or_else(|| Error::PathNotFound(path))?;
+                let mut commits: Vec<repo::OrderedCommit> =
+                    NonEmpty::flatten(subtree.to_nonempty()).into();
+                commits.sort_by(|commit, other| commit.id.cmp(&other.id));
+                commits.dedup_by(|commit, other| commit.id == other.id);
```
cloudhead, line 645: Sort and dedup by commit id? :thinking:

FintanH: So when I flatten the tree I get the same commits for multiple files. I might need to think about this more :thinking:

cloudhead: But what does sorting by commit id mean?

FintanH: Ah sorry, I misread your question. Sorting by the `Oid` so that the duplicates are beside each other, which is necessary for the `dedup`.

cloudhead: Ok, but don't we need the commits in a certain order for the file history? I must be missing something.

FintanH: Oh wait, sorry again hahaha. It's the "id" of the OrderedCommit, i.e. the result of calling enumerate on the revwalk -- the I'd signifying when we saw that commit on the walk. This is actually kind of irrelevant thanks to #111

PR #110, opened 2020-05-19
Clean up error types

FintanH:


src/file_system/error.rs
```diff
@@ -44,10 +44,10 @@ pub(crate) fn label_has_slash(item: &str) -> Error {
 pub enum Error {
     /// A `LabelError` specific error for parsing a
     /// [`Path`](super::path::Label).
-    #[error("Label error: {0}")]
+    #[error("malformed label: {0}")]
     Label(#[from] LabelError),
     /// A `PathError` specific error for parsing a [`Path`](super::path::Path).
-    #[error("Path error: {0}")]
+    #[error("malformed path: {0}")]
     Path(#[from] PathError),
```
CodeSandwich, line 51: The `PathError` is already `#[from]`, which implies that it's returned as `source`, so it shouldn't be printed it in the error message.

Same for `Label`.

src/file_system/error.rs
```diff
@@ -66,12 +66,12 @@ pub enum PathError {
 pub enum LabelError {
     /// An error signifying that a [`Label`](super::path::Label) is contains
     /// invalid UTF-8.
-    #[error("Label contains invalid UTF-8: {0}")]
+    #[error("label contains invalid UTF-8: {0}")]
     InvalidUTF8(String),
     /// An error signifying that a [`Label`](super::path::Label) contains a `/`.
-    #[error("Label contains a slash: {0}")]
+    #[error("label contains a slash: {0}")]
     ContainsSlash(String),
```
CodeSandwich, line 73: `anyhow` is much better than `String`, it's a valid `std::error::Error`, so it can be returned as a `source` and it's more elastic (it can contain any source, not only a `String`)

Same for `InvalidUTF8`.

src/vcs/git/error.rs
```diff
@@ -33,41 +33,38 @@ pub enum Error {
...
     /// exist as a tag. This could mean that the tag does not exist
     /// or that a branch or commit was provided by accident.
-    #[error("Provided tag name does not exist: {0}")]
+    #[error("provided tag name does not exist: {0}")]
     NotTag(TagName),
     /// A `revspec` was provided that could not be parsed into a branch, tag, or
     /// commit object.
-    #[error("Provided revspec could not be parsed into a git object: {0}")]
+    #[error("provided revspec could not be parsed into a git object: {0}")]
     RevParseFailure(String),
```
CodeSandwich, line 46: anyhow?

src/vcs/git/error.rs
```diff
@@ -33,41 +33,38 @@ pub enum Error {
...
     /// A [`str::Utf8Error`] error, which usually occurs when a git object's
     /// name is not in UTF-8 form and parsing of it as such fails.
-    #[error("Git object name is invalid UTF-8: {0}")]
+    #[error("git object name is invalid UTF-8: {0}")]
     Utf8Error(#[from] str::Utf8Error),
```
CodeSandwich, line 50: No need to print?

src/vcs/git/error.rs
```diff
@@ -78,13 +75,12 @@ impl From<diff::git::Error> for Error {
...
 pub(crate) enum TreeWalkError {
-    #[error("Entry is not a blob")]
+    #[error("entry is not a blob")]
     NotBlob,
-    #[error("Git object is a commit")]
+    #[error("git object is a commit")]
     Commit,
-    #[error("Git error: {0}")]
+    #[error("git error: {0}")]
     Git(Error),
```
CodeSandwich, line 84: Probably `#[from]` would be good here.

FintanH:
Thanks @CodeSandwich, geat call-outs :grin:

src/file_system/error.rs
FintanH: Hmmm so the point is that it's the underlying `String` for the `Label`, which isn't a valid label. So it's not so much a message but the data. Would I still use `anyhow::Error`? I feel like I wouldn't :thinking:

src/vcs/git/error.rs
FintanH: Same again here

src/file_system/error.rs
CodeSandwich: Ah, I haven't noticed that it's supposed to contain the invalid label value! I'm afraid that I may not be the only one who will get confused reading this code. How about we make it something like this?
```suggestion
    ContainsSlash {
        label_value: String
    },
```

src/vcs/git/error.rs
CodeSandwich: Right! Also same again :D

src/file_system/error.rs
FintanH: Great call :)

PR #111, opened 2020-05-19
Optimise Commit History

FintanH:
We can pass a pathspec to a diff_tree_to_tree so that we can focus on a single file.
Since we're focusing on a single file it means we don't need to keep
track of the whole Tree and instead just return the Vec of commits.
The final optimisation is just exit early on last_commit because that's
all we care about.

### DIFF
--- a/src/vcs/git/repo.rs
+++ b/src/vcs/git/repo.rs
@@ -20,13 +20,12 @@ use crate::{
     diff,
     diff::*,
     file_system,
-    tree::*,
     vcs,
     vcs::{git::error::*, VCS},
 };
 use git2::{BranchType, Oid};
 use nonempty::NonEmpty;
-use std::{cmp::Ordering, convert::TryFrom, str};
+use std::{convert::TryFrom, str};
 
 /// OrderedCommit is to allow for us to identify an ordering of commit history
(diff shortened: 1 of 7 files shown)


FintanH:
@cloudhead this simplified the `file-history` PR so I think if we're happy we can merge that into there, and then into `master` :shrug:

src/lib.rs
```diff
@@ -15,6 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+#![feature(test)]
```
FintanH, line 18: Note to self: remove

src/vcs/git/repo.rs
```diff
@@ -340,88 +340,66 @@ impl<'a> RepositoryRef<'a> {
...
         let old_tree = old_tree.map_or(Ok(None), |commit| commit.tree().map(Some))?;
 
-        let diff = self
-            .repo_ref
-            .diff_tree_to_tree(old_tree.as_ref(), Some(&new_tree), None)?;
+        let mut opts = git2::DiffOptions::new();
+        if let Some(path) = path {
+            opts.pathspec(path);
+        }
+        opts.skip_binary_check(true);
```
FintanH, line 398: move this into if

PR #115, opened 2020-05-22
Browser Default History

### LINKED ISSUE #114: Default Branch for Browser

FintanH:
Fixes #114 

* `VCS` now has `default_history_id` which means the backend should say where is the default place to defer to.
* We can define `default_history` in terms of `default_history_id` and `get_history` by default (pun not intended)
* `Browser` can now use `default_history` to kick off the history rather than taking `HEAD`.

### DIFF
--- a/src/vcs/git/repo.rs
+++ b/src/vcs/git/repo.rs
@@ -413,6 +413,11 @@ impl<'a> VCS<Commit, Error> for RepositoryRef<'a> {
     type HistoryId = String;
     type ArtefactId = Oid;
 
+    fn default_history_id(&self) -> Result<Self::HistoryId, Error> {
+        let reference = self.repo_ref.find_reference("refs/remotes/origin/HEAD")?;
+        Ok(reference.symbolic_target().unwrap().to_string())
+    }
+
     fn get_history(&self, history_id: Self::HistoryId) -> Result<History, Error> {
         self.revspec(&history_id)
     }
@@ -489,5 +494,22 @@ impl From<git2::Repository> for Repository {
 impl std::fmt::Debug for Repository {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, ".git")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use vcs::{git::error::Error, VCS};
+
+    #[test]
+    fn default_history_id() -> Result<(), Error> {
+        let repo = Repository::new("./data/git-platinum")?;
+        let repo = repo.as_ref();
+        assert_eq!(
+            repo.default_history_id(),
+            Ok("refs/remotes/origin/master".to_string())
+        );
+        Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


src/vcs/git/repo.rs
```diff
@@ -385,6 +385,11 @@ impl<'a> VCS<Commit, Error> for RepositoryRef<'a> {
     type HistoryId = String;
     type ArtefactId = Oid;
 
+    fn default_history_id(&self) -> Result<Self::HistoryId, Error> {
+        let reference = self.repo_ref.find_reference("refs/remotes/origin/HEAD")?;
```
FintanH, line 389: @kim Given my inability to understand how namespaces actually work under the hood, would this work?

kim: What do you mean, just follow the strings! 😂 

It's actually quite simple: there's everything which is normally under `refs/`, but prefixed by `refs/namespaces/{}`, where `{}` in `link` is `multibase(multihash(identity-hash))`.

Finding the default branch is a bit trickier. Normally, the server has a (virtually) immutable `HEAD` symref, which gets advertised when `git clone`ing (some details elided):

```
radicle-surf> git upload-pack --stateless-rpc --advertise-refs .
00fa86aa5860141e0e63c58e13380ef9afb381e05f71 HEADmulti_ack thin-pack side-band side-band-64k ofs-delta shallow deepen-since deepen-not deepen-relative no-progress include-tag multi_ack_detailed no-done symref=HEAD:refs/heads/master agent=git/2.26.2
003f86aa5860141e0e63c58e13380ef9afb381e05f71 refs/heads/master
0000
```

The `symref=HEAD:refs/heads/master` is what determines which branch will get checked out by default. If it does _not_ point to `master`, a "remote head" will be created in the local repo at `refs/remotes/<remote>/HEAD`, lest we forget (see also [git remote set-head](https://git-scm.com/docs/git-remote#Documentation/git-remote.txt-emset-headem)).

**However**, in `radicle-link`, we mandate that the default branch is determined by looking at the (signed) metadata -- because, it shallt be consistent for the entire project, regardless of what every replicated peer might say their `HEAD` is.

Unfortunately, `surf` is not aware of `link`, so we have two choices:

1. Initialise the `surf::Repository` with the (current) value of `default_branch`.

     The `Browser` would then ignore any `HEAD`s, and simply return what `refs/namespaces/<namespace>/refs/heads/<default_branch>` resolves to. `HEAD`s would still show up when listing branches, and every peer may choose to set it to a different value.

2. Rewrite all `HEAD`s  after validating the identity doc.

    This would mean to set `refs/namespaces/<namespace>/HEAD` _as well as_ `refs/namespaces/<namespace>/refs/remotes/**/HEAD`, overriding what was fetched from peers. The `Browser` would then look at `refs/namespaces/<namespace>/HEAD`, and if that doesn't exist, default to `refs/namespaces/<namespace>/refs/heads/master`.

The second option would be more gitiomatic, but I don't feel too comfortable about the rewriting business, as it can't be done atomically.

FintanH:
Closing in favour of #131

PR #116, opened 2020-05-26
Add conversion from &git2::Repository to RepositoryRef

kim:
Otherwise, we need to own a `git2::Repository` in the first place. The ref version would not be needed then, as we can just borrow `surf::Repository`.


kim:
Uuuh, `link` depends on a PR'd version now. Confusion arises. Thanks bye, nothing to see here.

PR #117, opened 2020-05-27
Browser Namespaces

### LINKED ISSUE #101: Support git namespaces
radicle-dev/radicle-link#144 will hand out surf `Browser`s on a per-namespace basis. That is, the `librad` user should not be able to surf the monorepo, but only the logical repo (namespace) it asked for.

While `libgit2` can set a namespace per-`git2::Repository`, this a) does not work as advertised ("This namespace affects all reference operations for the repo.") and b) is global state if we share the `git2::Repository` (see #100), which is problematic.

What could work is to construct a `Browser` with an optional namespace, and implement `list_branches` in terms of [`references`](https://docs.rs/git2/0.13.5/git2/struct.Repository.html#method.references) instead of `branches`.

FintanH:
Fixes #101 

Allows the caller to call `switch_namespace` so that they get back a new `Browser` that is set to the new `Namespace`. This means that `list_branches` and `list_tags` will be scoped to that `namespace`.

The main areas of interest are:
* `build.rs` sets up the namespace references for testing
* `vcs/git/reference/glob.rs` holds the logic for globbing and grabbing the references
* Our `Tag` type now consists of `Light` and `Annotated` so that we can add `TryFrom` for `git2::Reference`
* `vcs/git/reference.rs` was what I started with, but realised didn't quite need it. I think it could be a more interesting way of referring to a reference in the Browser though :thinking:


src/vcs/git/reference.rs
```diff
@@ -0,0 +1,224 @@
...
+                namespace: "moi".to_string(),
+                reference: Box::new(Ref::RemoteBranch {
+                    remote: "origin".to_string(),
+                    name: BranchName::new("master")
+                })
+            })
+        );
+
+        assert_eq!(
+            Ref::from_str("refs/namespaces/moi/refs/namespaces/toi/refs/tags/v1.0.0"),
```
kim, line 193: Holy... 🙀

src/vcs/git/reference/glob.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::vcs::git::{error, object::Namespace, repo::RepositoryRef};
+use std::fmt;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RefGlob {
+    /// When calling [`RefGlob::references`] this will return the references via
+    /// the glob `refs/tags/*`.
+    Tag,
+    /// When calling [`RefGlob::references`] this will return the references via
+    /// the glob `refs/heds/*`.
```
kim, line 27:
```suggestion
    /// the glob `refs/heads/*`.
```

src/vcs/git/reference/glob.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn from_branch_type(branch_type: git2::BranchType) -> Self {
+        match branch_type {
+            git2::BranchType::Remote => Self::RemoteBranch { remote: None },
+            git2::BranchType::Local => Self::LocalBranch,
+        }
+    }
+
+    pub fn references<'a>(&self, repo: &RepositoryRef<'a>) -> Result<References<'a>, error::Error> {
+        let namespace = repo
+            .which_namespace()?
```
kim, line 65: I see what you did there!

src/vcs/git/reference/glob.rs
```diff
@@ -0,0 +1,127 @@
...
+            Self::RemoteBranch { remote } => {
+                write!(f, "refs/remotes/")?;
+                match remote {
+                    None => write!(f, "**/*"),
+                    Some(remote) => write!(f, "{}/*", remote),
+                }
+            },
+            // Note: the glob below would be used, but libgit doesn't care for union globs.
+            // write!(f, "refs/{{remotes/**/*,heads/*}}")
+            Self::Branch => panic!(
```
kim, line 122: wait what? why is this panicking in `Display`?

FintanH: So it should be the above in the comment, but that glob doesn't work. It'll just return no references. So I'm not sure what's worse: 1) silent failure 2) fail as loud as possible. It's an internal module so the Display shouldn't be called on it. Maybe I should mark everything as `pub(super)`?

src/vcs/git.rs
```diff
@@ -171,6 +184,30 @@ impl<'a> Browser<'a> {
...
+    /// this `Browser`.
+    pub fn switch_namespace(self, namespace: &Namespace, revision: &str) -> Result<Self, Error> {
+        self.repository.switch_namespace(&namespace.to_string())?;
+        let history = self.repository.get_history(revision.to_string())?;
+        Ok(Browser {
+            snapshot: self.snapshot,
+            repository: self.repository,
+            history,
+        })
+    }
```
cloudhead, line 199: Doesn't it make more sense to have a `Browser::new_with_namespace` instead?

FintanH: I don't think a `new` version would make sense because we would take `self` to change the repository's namespace. It could be debated, however, whether this _needs_ to consume the `Browser` and give you back a new one or rather just take the reference and switch the namespace.

cloudhead: I'm not saying we should rename the function, but that it's more useful to create a new browser from a namespace, than to create a new browser and then create a new browser from an old browser just to switch the namespace!

FintanH: Ach so! Ya perfect idea :)

PR #120, opened 2020-05-29
Use heads rather than remotes

FintanH:
Bug fix

Using a remote wouldn't work downstream since the remotes wouldn't be set up when cargo grabs radicle-surf.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -68,7 +68,7 @@ fn main() {
     let golden_status = setup_namespace(
         curr_dir,
         "refs/namespaces/golden/refs/heads/master",
-        "refs/heads/dev",
+        "refs/heads/master",
     );
     assert!(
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -78,7 +78,7 @@ fn main() {
     let silver_status = setup_namespace(
         curr_dir,
         "refs/namespaces/golden/refs/namespaces/silver/refs/heads/master",
-        "refs/remotes/origin/dev",
```
FintanH, line 81: When cargo grabs surf as a dependency it doesn't have a notion of origin as a remote. But we do have `refs/heads/dev` so we use that instead.

massimiliano-mantione:
Looks good to me, and it fixes the issue :smiley:
Thanks! :heart:

Issue #124, opened 2020-06-03
Project Stats


FintanH:
Implement a function for the browser that gives the stats for some project. The stats include:
* Number of branches
* Number of commits - presumably the number of commits on the current branch you're browsing
* Number of contributors - should this number be based on the default branch (which needs to be fixed on #115) or on the current branch you're browsing?

cloudhead:
> Number of contributors - should this number be based on the default branch (which needs to be fixed on #115) or on the current branch you're browsing?

Branch you're browsing, same for commit count.

FintanH:
Ah, I just dropped a comment on the PR :smile: Cool that sounds good to me and makes things simpler

FintanH:
Closed via #125

PR #125, opened 2020-06-03
Project Stats

sarahscott:
- [x] `branch_count`
- [x] `commit_count`
- [x] `contributor_count`


src/vcs/git.rs
```diff
@@ -788,6 +789,55 @@ impl<'a> Browser<'a> {
...
+    /// assert_eq!(stats.branch_count, 5);
+    ///
+    /// // there should be 14 commits
+    /// assert_eq!(stats.commit_count, 14);
+    ///
+    /// // 4 contributors: Alexander Simmerl, Fintan Halpenny, FintanH, Rudolfs Ošiņš
+    /// assert_eq!(stats.contributor_count, 4);
+    ///
+    /// # Ok(())
+    /// # }
```
cloudhead, line 820: I think this kind of test makes more sense in a submodule as a unit test, rather than as a doctest, since it makes assumptions about the data available (`git-platinum`) and so might not make sense as documentation.

src/vcs/git.rs
```diff
@@ -788,6 +789,55 @@ impl<'a> Browser<'a> {
...
+        let history = self.repository.head()?;
+
+        let branch_count = self.list_branches(None)?.len();
+        let commit_count = history.iter().count();
+        let mut contributors: Vec<String> =
+            history.iter().cloned().map(|x| x.author.name).collect();
+        contributors.sort();
+        contributors.dedup();
+
+        let contributor_count = contributors.iter().count();
```
cloudhead, line 832: Since `contributors` is a `Vec` you can simply do `contributors.len()`

src/vcs/git/object.rs
```diff
@@ -126,6 +126,16 @@ impl BranchName {
     }
 }
 
+/// Stats for a repository
+pub struct Stats {
+    /// Number of commits
+    pub commit_count: usize,
+    /// Number of branches (local & remote)
```
cloudhead, line 133: I think we want to be using only local branches for this stat, otherwise it'll be confusing. You can do this with `    list_branches(Some(BranchType::Local))`

FintanH: Thanks, we weren't sure what would make sense so we defaulted to all :) But `Local` makes sense :+1:

src/vcs/git.rs
FintanH: All the doc tests/unit tests make assumptions about `git-platinum` data, though. The repository is treated as a set of "golden files".

src/vcs/git.rs
```diff
@@ -788,6 +789,55 @@ impl<'a> Browser<'a> {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn get_stats(&self) -> Result<Stats, Error> {
+        let history = self.repository.head()?;
+
+        let branch_count = self.list_branches(None)?.len();
+        let commit_count = history.iter().count();
+        let mut contributors: Vec<String> =
+            history.iter().cloned().map(|x| x.author.name).collect();
```
FintanH, line 828: Something that I think we need to think about is that this will change if the `Browser` switches branches. For example, if `master` is the default branch then we have all the contributors that have commits on that branch. But say there's another branch that is behind or ahead of master, then theoretically it will have less or more contributors.

FintanH: All good :ok_hand: https://github.com/radicle-dev/radicle-surf/issues/124#issuecomment-639332052

FintanH, line 828: This is more a tip for type inference. You can also do `collect::<Vec<_>>()` the `_` will be inferred and you're telling the compiler that you're expecting a `Vec`.

src/vcs/git.rs
cloudhead: Ah indeed!

src/vcs/git.rs
```diff
@@ -788,6 +789,55 @@ impl<'a> Browser<'a> {
...
+    /// assert_eq!(stats.contributor_count, 4);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn get_stats(&self) -> Result<Stats, Error> {
+        let history = self.repository.head()?;
+
+        let branch_count = self.list_branches(None)?.len();
+        let commit_count = history.iter().count();
```
FintanH, line 826: We can use `history.len()` here :)

sarahscott: I get this error when I try using `len()` on `history`:

```
vcs.rs(30, 1): method `len` not found for this
git.rs(826, 36): method not found in `vcs::History<vcs::git::object::Commit>`
```

FintanH: Two options:
1) Add len to History
2) Access the NonEmpty underneath by doing `history.0.len()` 

I think 1) could make more sense, and it'll basically be doing 2). You'll have to implement `empty` too because clippy will suggest it.

sarahscott: Sweet, will give this a go!

sarahscott: Clippy didn't suggest `empty` 😱  should I do it anyway?

sarahscott: Nevermind, clippy reared its head 😄  will add `is_empty`

src/vcs/git.rs
```diff
@@ -788,6 +789,59 @@ impl<'a> Browser<'a> {
...
+    /// assert_eq!(stats.commit_count, 14);
+    ///
+    /// // 4 contributors: Alexander Simmerl, Fintan Halpenny, FintanH, Rudolfs Ošiņš
+    /// assert_eq!(stats.contributor_count, 4);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn get_stats(&self) -> Result<Stats, Error> {
+        let history = self.repository.head()?;
```
FintanH, line 823: This shouldn't be needed

sarahscott: The `?` or the whole line?

FintanH: The whole line. Sorry, I should have been clearer. The user will already be on a branch they're browsing so there's no need to set the history to head.

FintanH:
Nice :) feel free to merge 🙌

PR #141, opened 2020-06-17
Serialize instances for `Diff`

cloudhead:


src/diff.rs
```diff
@@ -50,49 +58,97 @@ impl Default for Diff {
...
 #[derive(Debug, PartialEq, Eq)]
 pub struct Hunk {
     pub header: Line,
     pub lines: Vec<LineDiff>,
 }
 
 /// The content of a single line.
-pub type Line = Vec<u8>;
+#[derive(Debug, PartialEq, Eq)]
+pub struct Line(pub Vec<u8>);
```
xla, line 131: Why does it need to be pub? Thought it's an anti-pattern with newtypes.

src/diff.rs
```diff
@@ -125,26 +186,26 @@ impl From<LineDiffKind> for char {
 }
 
 impl LineDiff {
-    pub fn addition(line: Line, line_num: u32) -> Self {
+    pub fn addition(line: Vec<u8>, line_num: u32) -> Self {
```
xla, line 189: Can't we have `impl Into<Line>` and implement it for `Vec<u8>` on `Line`?

src/vcs/git.rs
```diff
@@ -1347,22 +1347,81 @@ mod tests {
...
+                    }
+                }]
+            };
+
+            let json = serde_json::json!({
+                "created": [],
+                "deleted": [],
+                "moved": [],
+                "copied": [],
+                "modified": [{
```
xla, line 1397: Can we use a diff which has content in these fields for a more thorough assertion?

src/diff.rs
cloudhead: We need it for serialization, and since there's no validation as to what can be a `Line`, I didn't see it as a problem. However, it should perhaps be `pub(crate)`!

src/vcs/git.rs
cloudhead: Contributors are welcome :joy:

FintanH: They really are

PR #148, opened 2020-07-08
Revision & Refactor

FintanH:
The aim of this PR is to get rid of the use of `str`s, because they're bad for types. Instead, we want to be explicit about what revision we're talking about.

We now have a struct called `Rev` which is composed of a `Ref` or an `Oid`. A `Ref` in turn can be a local/remote branch, a tag, or one of those nested in a namespace.

The API is pretty nice (if I do say so myself), where the `Browser` takes an `impl Into<Rev>` so we can use `Oid`, `Branch`, `TagName` and it does the conversion automagically.

The other refactor involves moving the definitions from `object.rs` into their own sub-modules. It was scaring me that `object` was a catch all for git things. I prefer having a hierarchy.


FintanH:
TODO: Setup remote in namespace and see if we can browse its history. This will emulate browsing a remote peer's branch

FintanH:
Note that this is a breaking change for upstream because we're getting rid of strings

src/vcs/git/ext.rs
```diff
@@ -0,0 +1,49 @@
...
+    let re = regex::Regex::new(r"(refs/namespaces/.*)*(refs/heads/|refs/remotes/|refs/tags/)(.*)")
+        .unwrap();
+
+    re.captures(spec)
+        .and_then(|c| c.get(3).map(|m| m.as_str()))
+        .ok_or(spec)
+}
+
+/// [`git2::Reference::is_tag`] just does a check for the prefix of `tags/`.
+/// This issue with that is, as soon as we're in 'namespaces' ref that
```
rudolfs, line 30:
```suggestion
/// The issue with that is, as soon as we're in 'namespaces' ref that
```

src/vcs/git/tag.rs
```diff
@@ -0,0 +1,160 @@
...
+        /// The name that references this `Tag`.
+        name: TagName,
+    },
+    /// An annotated git tag.
+    Annotated {
+        /// The Object ID for the `Tag`, i.e the SHA1 digest.
+        id: Oid,
+        /// The name that references this `Tag`.
+        name: TagName,
+        /// The named author of this `Tag`, if the `Tag` was annotated.
```
xla, line 80: How can the two fields be optional given that this is the `Annotated` variant?

src/vcs/git/stats.rs
```diff
@@ -0,0 +1,36 @@
+// This file is part of radicle-surf
```
xla, line 1: How did this end up in the PR? Was it moved?

src/vcs/git/namespace.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::{convert::TryFrom, fmt, str};
+
+/// A `Namespace` value allows us to switch the git namespace of
+/// [`super::Browser`].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Namespace {
+    /// Since namespaces can be nested we have a vector of strings.
+    /// This means that the namespaces `"foo/bar"` is represented as
+    /// `vec!["foo", "bar"]`.
+    pub(super) values: Vec<String>,
```
xla, line 28: `NonEmpty`? :shrug:

src/vcs/git/stats.rs
FintanH: Can you be more specific in what "it" is you're referring to?

src/vcs/git/tag.rs
FintanH: Apparently, they can still be `None`: https://docs.rs/git2/0.13.6/git2/struct.Tag.html#method.tagger + https://docs.rs/git2/0.13.6/git2/struct.Tag.html#method.message

FintanH:
> How much effort will it be to upgrade Upstream?

Shouldn't be much :+1: Can do an initial check with my local version.

FintanH:
Could you give it another going over? The other two changes were substantial enough

src/vcs/git/stats.rs
xla: The `Stats` seem unrelated to the change-seet.

FintanH: I moved them out of `object.rs` unchanged.

As I said in the description:
> The other refactor involves moving the definitions from object.rs into their own sub-modules. It was scaring me that object was a catch all for git things. I prefer having a hierarchy.

xla: Sorry didn't get that far, was too excited about the first three paragraphs. :service_dog:

PR #155, opened 2020-10-15
Update infra

FintanH:
Some of the way surf was set up was cribbing radicle-link. To keep this
theme up, we crib it some more and just have the one Dockerfile. I've
also updated the toolchain just to be on that bleeding edge!


xla (approved):
🕓 🆕 🤘 📅

FintanH:
Listing the references we find a reference called `refs/namespaces/golden/refs/remotes/kickflip/refs/heads/heelflip`, which I'm not sure where it's coming from. That third `refs` shouldn't be there.

rudolfs (approved):

PR #157, opened 2020-10-19
Fix backslash handling

rudolfs:
Added testcase.
Finto fixed it.


FintanH (approved):
Looks great, sick work :metal:

PR #158, opened 2020-10-19
Remote revisions

### LINKED ISSUE #153: Search for commit in remote branches
[`revision_branches`](https://github.com/radicle-dev/radicle-surf/blob/master/src/vcs/git/repo.rs#L261-L262) only searches for an `Oid` on the set of local branches, i.e. `refs/heads/*`. This is an issue if we wanted to search for a commit that exists only on a remote branch.

The fix here, I believe, is to update the above code to `RemoteBranch { remote: None }`. We should try to find a way to test this as well by having a remote branch that exists only as a remote.

FintanH:
Fixes #153 

The idea here is that we were only searching local branches for a particular commit. Will allow us to look at remote branches as well which is needed in Upstream since the only branches that exist after a new clone are remote branches.

### DIFF
--- a/src/vcs/git.rs
+++ b/src/vcs/git.rs
@@ -879,7 +879,10 @@ impl<'a> Browser<'a> {
     /// let branches = browser.revision_branches(Oid::from_str("27acd68c7504755aa11023300890bb85bbd69d45")?)?;
     /// assert_eq!(
     ///     branches,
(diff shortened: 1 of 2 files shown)


rudolfs (approved):

xla (approved):
🎎 📏 🚣 ✂️

PR #160, opened 2020-11-14
Use git ext

FintanH:
The parsing of references in radicle-surf was quite ad-hoc and used
regexes. We wish to move to a more disciplined approach by using
radicle-git-ext's family of reference structures.

Admittedly, there's still some ad-hoc parsing or removing of prefixes
but it's still better than before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "radicle-surf"
 description = "A code surfing library for VCS file systems"
 readme = "README.md"
-version = "0.5.0"
+version = "1.0.0"
 authors = ["The Radicle Team <dev@radicle.xyz>"]
 edition = "2018"
 homepage = "https://github.com/radicle-dev/radicle-surf"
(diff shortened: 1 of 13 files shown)


FintanH:
I think I'd like to move `radicle-git-ext` into its own publishable crate so that we can have lax version bounds on surf¸ link, and proxy. What do you guys think?

FintanH:
> I think I'd like to move radicle-git-ext into its own publishable crate so that we can have lax version bounds on surf¸ link, and proxy. What do you guys think?

Looking at migrating it to its own repository, there are two things:
* `radicle-std-ext` would need to be migrated too
* `radicle-git-ext` uses `librad-test`'s `json_roundtrip` -- since this isn't `librad` specific we could also create a testing crate

kim:
I can't say I'm keen on this - neither of the `*-ext` crates is in a state where I'd be happy to stabilise them. If this is only about reference names, I would prefer to maintain a _copy_ of those types + macros for surf over the nightmare that is `cargo` dependency resolution.

FintanH:
Ok, I'll copy them over and create a ticket to do a crate migration at some point in the future :+1:

FintanH:
The duplication has it's own issue while I was carving this out. Gonna drop this until another time and find a different fix.

PR #161, opened 2020-11-17
Use nom for parsing references

FintanH:
The regex had a subtle bug in it that I wasn't willing to put the effort
in to figure out why. Instead I've switched to use the nom parsing
library so I can have a better understanding of the semantics. However,
I also did not put in the effort to figure out how to make the errors
better. Any failure is considered a "MalformedRef".

### DIFF
--- a/src/vcs/git/reference.rs
+++ b/src/vcs/git/reference.rs
@@ -173,8 +173,6 @@ pub mod parser {
     pub fn namespace(s: &str) -> Result<Ref, Error> {
         bytes::complete::tag(NAMESPACES)(s).and_then(|(rest, _)| {
             component(rest).and_then(|(rest, namespace)| {
-                println!("REST: {}", rest);
-                println!("namespace: {}", namespace);
                 Ok(Ref::Namespace {
                     namespace: namespace.to_owned(),
                     reference: Box::new(parse(rest)?),
--- a/src/vcs/git/reference.rs
+++ b/src/vcs/git/reference.rs
@@ -15,10 +15,11 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::vcs::git::{repo::RepositoryRef, BranchName, Namespace, TagName};
-use regex::Regex;
 use std::{fmt, str};
+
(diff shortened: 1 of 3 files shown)


FintanH:
@NunoAlexandre tagging you because you played with `nom` and thought you might have some good feedback on usage. The library is a bit of a war ground with `str` and `[u8]`.

NunoAlexandre (commented):
nomnom :cookie: 

Looks good to me :)

xla (approved):
utACK

💧 🅿️ 👕 🍤

src/vcs/git/reference.rs
```diff
@@ -119,68 +120,82 @@ impl fmt::Display for Ref {
...
+        bytes::complete::tag(TAGS)(s).map(|(name, _)| Ref::Tag {
+            name: TagName::new(name),
+        })
+    }
+
+    pub fn namespace(s: &str) -> Result<Ref, Error> {
+        bytes::complete::tag(NAMESPACES)(s).and_then(|(rest, _)| {
+            component(rest).and_then(|(rest, namespace)| {
+                println!("REST: {}", rest);
+                println!("namespace: {}", namespace);
```
kim, line 177: Yes?

kim (approved):
lgtm

What was [the issue](https://github.com/radicle-dev/radicle-surf/pull/160#issuecomment-728797890)?

src/vcs/git/reference.rs
FintanH: :flushed:

FintanH:
> What was the issue?

I got into all kinds of annoyances when `PeerId` got involved as a remote -- using `TryFrom` -- and then the mixed-use of `librad::RefLike` and `surf::RefLike` was too much of a headache. I couldn't see it being worth the maintenance.

kim:
I see. We could consider releasing only the `git_ext::reference::name` module as
a separate crate. Maybe "rightpad" would be an appropriate name :thinking:

FintanH:
Hahaha I'm ok with releasing something more complete and polished. It would be good to know what you're thinking needs work in that area and I could chip in with getting it publishable :)

PR #165, opened 2020-11-24
Bump to v0.5.2

FintanH:


rudolfs (approved):

xla (approved):
😛 🎯 🏝 📢

FintanH:
Might need to clear the cache of this machine because it has the existing reference that we're not deleting now.

PR #173, opened 2021-03-29
Source Browsing

FintanH:


source/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = ["The Radicle Team <dev@radicle.xyz>"]
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[dependencies]
+log = "0.4"
+lazy_static = "1.4"
+nonempty = "0.6"
+serde = { version = "1.0", features = [ "derive" ] }
+syntect = "4.2"
```
cloudhead, line 13: I wonder if this (surf) is the right place for syntax highlighting. Perhaps it is :shrug: but maybe behind a feature flag then?

FintanH: Feature flag works for me :ok_hand:

xla (approved):
📝 💒 🚄 🌼

PR #175, opened 2021-04-01
rework the radicle-source module

FintanH:
To tidy house, we split radicle-source into lots of separate component
modules so that it's easier to browse (weh weh) and change related code.

We also put the syntax highlighting behind a "syntax" feature flag.

### DIFF
--- a/source/Cargo.toml
+++ b/source/Cargo.toml
@@ -5,13 +5,21 @@ authors = ["The Radicle Team <dev@radicle.xyz>"]
 edition = "2018"
 license = "GPL-3.0-or-later"
 
+[features]
+syntax = ["syntect"]
+
 [dependencies]
 log = "0.4"
 lazy_static = "1.4"
 nonempty = "0.6"
 serde = { version = "1.0", features = [ "derive" ] }
-syntect = "4.2"
+syntect = { version = "4.2", optional = true }
 thiserror = "1.0"
+
+[dependencies.git2]
+version = ">= 0.12"
+default-features = false
+features = []
 
 [dependencies.radicle-surf]
 path = "../surf"
--- a/source/src/branch.rs
+++ b/source/src/branch.rs
@@ -1,0 +1,105 @@
+// This file is part of radicle-surf
+// <https://github.com/radicle-dev/radicle-surf>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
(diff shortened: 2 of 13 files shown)


source/src/lib.rs
```diff
@@ -17,938 +17,31 @@
...
-where
-    P: ToString,
-{
-    identifier.map_or(BranchType::Local, |identifier| BranchType::Remote {
-        // We qualify the remotes as the PeerId + heads, otherwise we would grab the tags too.
-        name: Some(format!("{}/heads", identifier.to_string())),
-    })
-}
+pub mod tag;
+pub use tag::{tags, Tag};
```
geigerzaehler, line 47: Tag `exports` nothing besides `tags` and `Tag`. To avoid ambiguity and keep the public API small I’d suggest to not publicly expose the `tag` module. This also applies to other modules.

In general, I feel that exposing everything at the crate level and not exposing any modules would be manageable given the size of the crate and would make it more easily digestible.

FintanH: My thoughts are that it gives flexibility of how an upstream caller imports things from `radicle_source`. If they have their own version of `Tag` or `tags` they can import `radicle_source::tag` and so they're now scoped to `tag::Tag` and `tag::tags`. I suppose they could also scope it to `radicle_source::Tag` and `radicle_source::tags` :thinking: 

In the end, there's nothing to hide behind `mod tag` and so there's no need to make it private either.

What are the cons of exposing modules?

geigerzaehler: > If they have their own version of `Tag` or `tags` they can import `radicle_source::tag` and so they're now scoped to `tag::Tag` and `tag::tags`.

That’s possible but likely confusing since now `tag` (from `radicle_source`) is totally disconnected from `Tag` or `tags` which come from a different module. In that case it’s more likely to use `radicle_source::Tag` as you noted.

> What are the cons of exposing modules?

As I mentioned there are two donwsides. One is that it enables two ways to import things. That may not be a huge issue though. It becomes more annoying if some parts of the module are re-exported at the top-level and some parts aren’t and you need to look up which. The second downside is that it makes it harder to grasp the API through the docs. Instead of seeing everything in the crate page I have to go down into the modules. And those items that are re-exported are listed twice in the documentation which leads to more noise.

Another point worth mentioning is that it’s harder to control the public API when the modules are public. If all modules are private then a glance at `src/lib.rs` tells me what is part of the public API. Otherwise it’s easier for code to sneak in to a module that is only intended for internal use and have `pub(crate)` visibility.

FintanH: I think I have a compromise :) Instead of top-level re-exports of these types we organise them into logical modules that re-export them there. So I already introduced `object/*`. For `tag`, `branch`, and `commit` I'm thinking something like `reference`.

How do you feel about this?

cloudhead: Some crates have a `prelude` module that exports the most used types, intended to be used like eg. `use radicle_surf::prelude::*` :shrug:

cloudhead (dismissed):

source/src/lib.rs
FintanH: Actually, also on this note: I'd like to de-duplicate some of these types anyways because they can just come from `radicle-surf` in the first place.

geigerzaehler: > I think I have a compromise :)

No need for a compromise. Go with what you feel is right! I just wanted to add some perspective. My comments should not prevent this from getting merged.

> Instead of top-level re-exports of these types we organise them into logical modules that re-export them there.

Wouldn’t this make it even harder to track what is exported and from where?

FintanH: Thanks @geigerzaehler, I'll have another pass at this when I de-duplicate the types :)

Issue #176, opened 2021-04-06
Testing strategy


FintanH:
After reading this article https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html, I'm thinking that some of the testing organisation could be changed.

For example, the current layout is generally a `mod test { /* test cases */ }`. In the article, it's recommended to use `mod test` with a `test.rs` file so that the compiler doesn't think the original module itself has changed.

There are a lot of doc tests, which the article says are slow, but I think since it provides a good view of the API and what the functions do they should stay.

A set of integration tests might be a nice addition, but no particular plan in what flows could be tested.

uint:
I'm new and only starting to dig into this project so take me with a grain of salt, but would it make sense to also revisit the whole thing with having to set up test fixtures (git refs) by external script before running `cargo test`?

It doesn't feel right that I can't clone this project, run `cargo test` and have it Just Work©. To me, in an ideal world the tests themselves would ensure the right environment is there (as much as possible).

FintanH:
Thanks for taking the time to look around @uint :blush: 

> but would it make sense to also revisit the whole thing with having to set up test fixtures (git refs) by external script before running cargo test?

I'm definitely for discussing this! The fixtures testing has always left something to be desired. I tried to make it as seamless as possible by using a `build.rs` file to set up the submodule. That became too much of a headache. For whatever reason, it wouldn't update the refs properly and I ended up having to incorporate an env var. So I just ditched that method.

What kind of approach were you thinking of? Just keep in mind that it needs to work on the unit tests as well as the doc tests, and I'd hope to not have any noise in the doc tests, if possible.

uint:
@FintanH I'm guessing since you're using submodules to keep that git repo around, it's not really possible/straightforward to keep other refs in it, right? Git probably just checks out one tree and that's it, and that's why you add other refs dynamically.

Have you considered ditching submodules and packaging up the whole git repo in a tarball, along with all the refs needed for tests? Then the only thing tests need to do for prep is untar the thing to some tempdir. I'm not sure about doctests, but if we have a function `ensure_git_platinum()` or something that does the prep, we could include [that line in a doc example and hide it](https://doc.rust-lang.org/rustdoc/documentation-tests.html). I guess that still creates some noise in the source code.

FintanH:
[...]

PR #180, opened 2021-04-19
Handle test assets without submodules and extra setup

uint:
This is part of the work for #176, but not all of it. It's just my proposal! We can scrap it if it doesn't work for you.

## How it works

`git-platinum` is now kept as a tarball in `data/git-platinum.tgz`. All the necessary refs are in there already. There's a `build.rs` script that extracts the archive into `data/git-platinum`, which is ignored by git (git doesn't like nested repos, so we don't want to try checking it in).

## Updating `git-platinum`
There's a script for that - `scripts/update-git-platinum.sh`. It will fetch the newest version of [git-platinum](https://github.com/radicle-dev/git-platinum), add the necessary refs, then update and commit the archive.

## The big win
It should now be possible to simply `git clone` the repo and run `cargo test` without extra setup necessary, keeping it simple for future contributors. Also no more submodules.

## Drawbacks
* 2 build dependencies added - `flate2` and `tar`.
* The build script doesn't distinguish between test builds and other builds. It will unpack the archive even if we're not running any tests. Any errors there will stop the build.


FintanH:
> The build script doesn't distinguish between test builds and other builds. It will unpack the archive even if we're not running any tests. Any errors there will stop the build.

I thought we might have been able to use:
```rust
if cfg!(test) {
  unpack
  println!
}
```

But it seems like the `build.rs` file doesn't know about `cargo test` being run and so it never fires. Not a big deal anyway since it'll only run once unless the tar file is changed. If it fails the person won't be able to run the tests which would be bad anyway and they should report an issue.

scripts/update-git-platinum.sh
```diff
@@ -0,0 +1,36 @@
...
+tar -czf git-platinum.tgz git-platinum
+mv git-platinum.tgz $TARBALL_PATH
+
+# Clean up.
+rm -rf git-platinum
+
+# Commit.
+git reset
+git add $TARBALL_PATH
+git commit -m "Update git-platinum"
```
FintanH, line 36: I'd prefer that the person performing this intentionally commits, just in case they accidentally updated or something along those lines.

FintanH (commented):
I'm in favour of this! What would the rest of the work look like? :)

uint:
Rebased and added sign-offs.

scripts/update-git-platinum.sh
uint: Yeah, I wasn't really sure if that was a good idea. Removed now!

scripts/setup-branches.sh
```diff
@@ -11,5 +11,5 @@ while IFS= read -r line
 do
```
FintanH, line 11: We don't need this file anymore, right? We just use the same logic in the update script afaict. Would you mind removing this if that's the case?

data/README.md
```diff
@@ -0,0 +1,3 @@
+# Updating [git-platinum](https://github.com/radicle-dev/git-platinum)
+Just `cd` into the repo root, then run `scripts/update-git-platinum.sh`.
+This will update the tarball and commit the change.
```
FintanH, line 3: We need to change this instruction now to tell the person to commit the change :) I'd also suggest a template:

```suggestion
This will update the tarball and it's necessary to commit the change. We provide a template below so that we can easily identify changes to `git-platinum`. Please fill in the details that follow a comment (`#`):


    data/git-platinum: # short reason for updating

    # provide a longer reason for making changes to git-platinum
    # as well as what has changed.

```

scripts/update-git-platinum.sh
```diff
@@ -0,0 +1,31 @@
...
+set -euo pipefail
+
+## This is an update script for /data/git-platinum. It's meant to be executed
+## from the root of the repo.
+
+TARBALL_PATH=data/git-platinum.tgz
+
+# This is here in case the last script run failed and it never cleaned up.
+# Is there a better way to handle this?
+rm -rf git-platinum
```
FintanH, line 11: Maybe a safer thing to do here is checking if the directory exists and asking the person to remove it and exiting the script.

I'm worried that we'd remove a `git-platinum` by accident, e.g. we ran the script in the wrong directory :sweat_smile:

scripts/update-git-platinum.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+## This is an update script for /data/git-platinum. It's meant to be executed
+## from the root of the repo.
```
FintanH, line 5: We could check that we're at the root by doing:

```bash
BASE=$(basename pwd)

if [ "${BASE}" -ne "radicle-surf" ]
then
   echo "this script should be run from the root of radicle-surf"
   exit 1
fi
```

Or something along those lines :)

surf/README.md
```diff
@@ -11,8 +11,8 @@ Alongside this, we will wish to take two snapshots and view their differences.
 
 ## Contributing
 
-To get started on contributing you can check out our [developing guide](./DEVELOPMENT.md), and also
-our [LICENSE](./LICENSE) file.
+To get started on contributing you can check out our [developing guide](../DEVELOPMENT.md), and also
```
FintanH, line 14: Thanks for catching these :)

FintanH (changes requested):
Just a few things to clean up and then this is good to be merged :)

scripts/update-git-platinum.sh
uint: What if the whole script worked in some sort of `.workdir` and we added that one to `.gitignore`?

scripts/setup-branches.sh
uint: Yep, I meant to do that! Probably forgot. I'll add the change shortly.

scripts/update-git-platinum.sh
uint: That's excellent. I'll give it a try. I was wondering if that's possible, but my bash powers are honestly fairly weak ;)

FintanH: Mine aren't much better, but I sure can google! :P

scripts/update-git-platinum.sh
FintanH: https://github.com/radicle-dev/radicle-surf/pull/180#discussion_r618214499 would be enough to resolve this I think :)

uint: I still created a sort of workdir. Feels cleaner anyway. How's that look?

data/README.md
```diff
@@ -0,0 +1,9 @@
+# Updating [git-platinum](https://github.com/radicle-dev/git-platinum)
+Just `cd` into the repo root, then run `scripts/update-git-platinum.sh`.
+This will update the tarball and it's necessary to commit the change. We provide a template below so that we can easily identify changes to `git-platinum`. Please fill in the details that follow a comment (`#`):
```
FintanH, line 3: Not to be a pain (but being a pain nonetheless): can we format the column size to 80? :)

scripts/update-git-platinum.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Verify that the script is run from project root.
+BASE=$(basename $(pwd))
+
+if [ "${BASE}" != "radicle-surf" ]
+then
+   echo "this script should be run from the root of radicle-surf"
```
FintanH, line 9:
```suggestion
   echo "ERROR: this script should be run from the root of radicle-surf"
```

Just to make it clear :smile:

FintanH (approved):
Looks good! Just some small suggestions and then we're _really_ ready to merge :)

data/README.md
uint: No worries, good point.

FintanH:
Merged via https://github.com/radicle-dev/radicle-surf/commit/f238a450c7cbee4faad1f663871716fff3b29855

PR #181, opened 2021-04-24
`BranchSelector` enum

### LINKED ISSUE #142: List Branches Enum
To choose which branches we want to list we pass in an `Option<BranchType>`. Passing in `None` can be confusing. A better way to go about this would be to have an `enum BranchSelector` which covers the 3 cases of: all, local, and remote branches.

The default case will be: all.

therealyingtong:
Closes #142.

### DIFF
--- a/source/src/revision.rs
+++ b/source/src/revision.rs
@@ -113,7 +113,7 @@ pub fn remote<P, U>(
 where
     P: Clone + ToString,
 {
-    let remote_branches = branches(browser, into_branch_selector(Some(peer_id.clone())))?;
+    let remote_branches = branches(browser, Some(peer_id.clone()).into())?;
     Ok(
         NonEmpty::from_vec(remote_branches).map(|branches| Revisions {
             peer_id,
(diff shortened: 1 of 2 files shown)


FintanH:
Hey @therealyingtong :wave: thanks for the contribution! :blush: As per our [DCO](https://github.com/radicle-dev/radicle-surf/blob/master/DCO), we require that you [sign-off](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s) all your commits. Let me know if you need help with this :)

source/src/revision.rs
```diff
@@ -113,7 +113,7 @@ pub fn remote<P, U>(
 where
     P: Clone + ToString,
 {
-    let remote_branches = branches(browser, Some(into_branch_type(Some(peer_id.clone()))))?;
+    let remote_branches = branches(browser, into_branch_selector(Some(peer_id.clone())))?;
```
cloudhead, line 116: While we're at it, can we turn the `into_branch_selector` function into a `From<Option<PeerId>> for BranchSelector` instance?

FintanH: Ya good point, although it won't be – technically – PeerId, it'll be P

therealyingtong:
Might a similar refactor (`Option<BranchType> -> BranchSelector`) be valuable in `radicle-upstream`?

https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/coco/src/source.rs#L552

FintanH (approved):

FintanH:
> Might a similar refactor (Option<BranchType> -> BranchSelector) be valuable in radicle-upstream?

That's not necessary. This code is currently duplicated while we move some things around. This PR https://github.com/radicle-dev/radicle-upstream/pull/1753 will resolve this for us.

FintanH:
Thanks @therealyingtong! :blush:

PR #185, opened 2021-05-04
surf: build infrastructure

FintanH:
When we attempted to publish the new version of `radicle-surf` we ran
into issues with the `build.rs` file executing -- it was missing the
path to `data/git-platinum.tgz` in its isolated environment.

The fix here is to move the `data` artifacts into the `surf` directory
and add the `git-platinum.tgz` file as an `include` artifact.
When the `build.rs` file executes it will detect whether it ran with
`cargo build` or `cargo publish` by looking at the current directory.
Depending on which it is using it will unpack the tar file into the
correct directory.

Moving the `data` directory required us to also change the tests to
point to the correct directory.


FintanH:
There was a few occasions clippy was giving out to me :thinking:

xla (approved):
🍂 🗾 🚍 💿

PR #188, opened 2021-05-11
surf: fix scripts/update-git-platinum.sh

geigerzaehler:
The script stopped working when `data` was moved to `surf`. We also
provide more comprehensive documentation on how to use the script.


FintanH (dismissed):
Awesome, thanks for catching this!

PR #189, opened 2021-05-11
surf: add tests for annotated tags

geigerzaehler:
This also includes an update to git-platinum where we add an annotated tag.


FintanH (dismissed):
Looks good :+1: fyi clippy is complaining

FintanH (approved):

PR #194, opened 2021-05-13
surf: add remote case

FintanH:
We add a case for a remote branch that is of the form
`refs/remotes/banana/orange/pineapple`. We confirm that listing the
branches will return `orange/pineapple`.


xla (approved):
💮 🏅 🚐 🛄

PR #195, opened 2021-05-17
surf: Add failing tests for branch with slashes

geigerzaehler:


FintanH:
@geigerzaehler actually you were just missing an expectation case for the test :grimacing: Everything seems to work fine, so perhaps it is the namespacing that's causing the issue you see?

geigerzaehler:
You’re right :see_no_evil:. I don’t know what caused the issue when I was working on it. Maybe I didn’t update the repo refs correctly. I was also able to track down some [definitive documentation](https://libgit2.org/libgit2/#HEAD/group/reference/git_reference_foreach_glob) on globs which confirms that `*` matches any sequence of characters including `/`.

FintanH:
Well, let's at least rebase this and merge it so we have these test cases :) If you want to try adding a case using a namespace and listing a particular remote that should make sure that our bases are covered. Otherwise, I guess I'll wait for you to find this mysterious bug again :P

geigerzaehler:
And this is already covered by the tests :facepalm:: https://github.com/radicle-dev/radicle-surf/blob/eff4cf3d9685f9eb051c864c00b62bbeb21da7a9/surf/src/vcs/git.rs#L1218-L1221

PR #196, opened 2021-05-18
surf: Add the ability to list remote tags

geigerzaehler:
`Browser::list_tags` now accepts a `RefSelector` that allows users to list tags from remotes. This is similar to the behavior of `Browser::list_branches`.


surf/src/vcs/git.rs
```diff
@@ -675,23 +682,35 @@ impl<'a> Browser<'a> {
...
+    ///     },
+    /// ]);
+    /// let golden = golden.switch_namespace(&Namespace::try_from("golden")?, Branch::local("master"))?;
+    ///
+    /// let branches = golden.list_tags(RefScope::Remote { name: Some("kickflip".to_string()) })?;
+    /// assert_eq!(branches, vec![
+    ///     Tag::Light {
+    ///         id: Oid::from_str("d3464e33d75c75c99bfb90fa2e9d16efc0b7d0e3")?,
+    ///         name: TagName::new("kickflip/v0.1.0"),
+    ///         remote: None,
```
FintanH, line 705: Shouldn't this be:
```rust
name: TagName::new("v0.1.0"),
remote: Some("kickflip"),
```

surf/src/vcs/git/reference/glob.rs
```diff
@@ -37,22 +37,19 @@ pub enum RefGlob {
         remote: Option<String>,
     },
     /// When calling [`RefGlob::references`] this will return the references via
-    /// the globs `refs/heads/*` and `refs/remotes/**/*`.
-    Branch,
+    /// the glob `refs/tags/*`.
```
FintanH, line 40: Documentation here needs updating

surf/src/vcs/git/tag.rs
```diff
@@ -155,6 +158,7 @@ impl<'repo> TryFrom<git2::Reference<'repo>> for Tag {
                     Ok(Tag::Light {
                         id: commit.id(),
                         name,
+                        remote: None,
```
FintanH, line 161: AH, I guess this explains the doctest I commented on above :sweat_smile:

surf/src/vcs/git.rs
```diff
@@ -121,28 +121,29 @@ impl From<git2::Buf> for Signature {
     }
 }
 
-/// Select the branches to be listed.
-pub enum BranchSelector {
+/// Determines whether to look for local or remote references or both.
+pub enum RefScope {
```
FintanH, line 125: Hmmmm maybe `RefScope` should now take the category it's interested in, i.e. `head`, `tags`, `notes`, etc. What do you think?

I say this because the `From` impl defaults to `heads`, which I think you've mentioned may be a problem

surf/src/vcs/git/repo.rs
```diff
@@ -107,8 +112,13 @@ impl<'a> RepositoryRef<'a> {
     /// # Errors
     ///
     /// * [`Error::Git`]
-    pub fn list_tags(&self) -> Result<Vec<Tag>, Error> {
-        RefGlob::Tag
+    pub fn list_tags(&self, scope: RefScope) -> Result<Vec<Tag>, Error> {
+        let ref_glob = match scope {
```
FintanH, line 116: Can we move these two `match`es into functions on `RefGlob`? i.e. `RefGlob::tag(scope)` and `RefGlob::branch(scope)`

FintanH (changes requested):

surf/src/vcs/git.rs
geigerzaehler: This is resolved now

surf/src/vcs/git.rs
geigerzaehler: The `From` implementation that hard-codes `heads` is indeed a problem. But adding the category as a fieldseems to add some redundancy. `RefScope` is passed to `list_branches()` and `list_tags()` which already determines the category.

To address the issue we could remove the `From` instance and move that code to it’s [single call-site in upstream](https://github.com/radicle-dev/radicle-upstream/blob/8a6413549f9bd37bbaafbb77eb3de09853e83955/proxy/api/src/http/source.rs#L165).

surf/src/vcs/git/repo.rs
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-surf/pull/196/commits/639b54cd3cc3abd1bafec8381edc45db7d569d87

surf/src/vcs/git.rs
FintanH: Ya that probably makes the most sense, good shout :+1:

FintanH (dismissed):
Nice one!

PR #201, opened 2022-02-15
Add `EofNewLine` enum on `ModifiedFile`

xphoniex:
Currently we're getting diffs like this when EOF is missing new line:
```
diff --git a/HomebrewFormula b/HomebrewFormula
index 4ee8025..1ffaf04 120000
--- a/HomebrewFormula
+++ b/HomebrewFormula
@@ -1 +1 @@
-pkg/brew/radicle-tools.rb
\ No newline at end of file
+pkg/brew
\ No newline at end of file
```

So we remove the "\ No newline at end of file" message and a flag to Hunk instead.

### DIFF
--- a/surf/src/diff.rs
+++ b/surf/src/diff.rs
@@ -94,9 +94,22 @@ pub struct CopyFile {
     serde(rename_all = "camelCase")
 )]
 #[derive(Clone, Debug, PartialEq, Eq)]
+pub enum EofNewLine {
+    OldMissing,
+    NewMissing,
+    BothMissing,
+}
+
+#[cfg_attr(
+    feature = "serialize",
+    derive(Serialize),
+    serde(rename_all = "camelCase")
+)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ModifiedFile {
     pub path: Path,
     pub diff: FileDiff,
+    pub eof: Option<EofNewLine>,
 }
 
 /// A set of changes belonging to one file.
@@ -274,7 +287,7 @@ impl Diff {
                                     let mut path = parent_path.borrow().clone();
                                     path.push(new_file_name.clone());
 
-                                    diff.add_modified_file(path, vec![]);
+                                    diff.add_modified_file(path, vec![], None);
                                 }
                                 old_entry_opt = old_iter.next();
                                 new_entry_opt = new_iter.next();
@@ -403,13 +416,19 @@ impl Diff {
         Ok(())
     }
 
-    pub(crate) fn add_modified_file(&mut self, path: Path, hunks: Vec<Hunk>) {
+    pub(crate) fn add_modified_file(
+        &mut self,
+        path: Path,
+        hunks: Vec<Hunk>,
+        eof: Option<EofNewLine>,
+    ) {
         // TODO: file diff can be calculated at this point
         // Use pijul's transaction diff as an inspiration?
         // https://nest.pijul.com/pijul_org/pijul:master/1468b7281a6f3785e9#anesp4Qdq3V
         self.modified.push(ModifiedFile {
             path,
             diff: FileDiff::Plain { hunks },
+            eof,
         });
     }
 
@@ -425,6 +444,7 @@ impl Diff {
         self.modified.push(ModifiedFile {
             path,
             diff: FileDiff::Binary,
+            eof: None,
         });
     }
 
@@ -544,6 +564,7 @@ mod tests {
             modified: vec![ModifiedFile {
                 path: Path::with_root(&[unsound::label::new("banana.rs")]),
                 diff: FileDiff::Plain { hunks: vec![] },
+                eof: None,
             }],
         };
 
@@ -629,6 +650,7 @@ mod tests {
                     unsound::label::new("banana.rs"),
                 ]),
                 diff: FileDiff::Plain { hunks: vec![] },
+                eof: None,
             }],
         };
 
--- a/surf/src/diff/git.rs
+++ b/surf/src/diff/git.rs
@@ -20,7 +20,7 @@ use std::convert::TryFrom;
 use thiserror::Error;
 
 use crate::{
-    diff::{self, Diff, Hunk, Line, LineDiff},
+    diff::{self, Diff, EofNewLine, Hunk, Line, LineDiff},
     file_system::Path,
     vcs,
 };
@@ -89,6 +89,8 @@ impl<'a> TryFrom<git2::Diff<'a>> for Diff {
 
                     if let Some(patch) = patch {
                         let mut hunks: Vec<Hunk> = Vec::new();
+                        let mut old_missing_eof = false;
+                        let mut new_missing_eof = false;
 
                         for h in 0..patch.num_hunks() {
                             let (hunk, hunk_lines) = patch.hunk(h)?;
@@ -97,12 +99,34 @@ impl<'a> TryFrom<git2::Diff<'a>> for Diff {
 
                             for l in 0..hunk_lines {
                                 let line = patch.line_in_hunk(h, l)?;
+                                match line.origin_value() {
+                                    git2::DiffLineType::ContextEOFNL => {
+                                        new_missing_eof = true;
+                                        old_missing_eof = true;
+                                        continue;
+                                    },
+                                    git2::DiffLineType::AddEOFNL => {
+                                        old_missing_eof = true;
+                                        continue;
+                                    },
+                                    git2::DiffLineType::DeleteEOFNL => {
+                                        new_missing_eof = true;
+                                        continue;
+                                    },
+                                    _ => {},
+                                }
                                 let line = LineDiff::try_from(line)?;
                                 lines.push(line);
                             }
                             hunks.push(Hunk { header, lines });
                         }
-                        diff.add_modified_file(path, hunks);
+                        let eof = match (old_missing_eof, new_missing_eof) {
+                            (true, true) => Some(EofNewLine::BothMissing),
+                            (true, false) => Some(EofNewLine::OldMissing),
+                            (false, true) => Some(EofNewLine::NewMissing),
+                            (false, false) => None,
+                        };
+                        diff.add_modified_file(path, hunks, eof);
                     } else if diff_file.is_binary() {
                         diff.add_modified_binary_file(path);
                     } else {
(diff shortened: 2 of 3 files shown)


surf/src/diff.rs
```diff
@@ -124,6 +124,8 @@ pub enum FileDiff {
 pub struct Hunk {
     pub header: Line,
     pub lines: Vec<LineDiff>,
+    pub old_missing_eof: bool,
+    pub new_missing_eof: bool,
```
cloudhead, line 128: Shouldn't this be on the file? There can only be one missing eof per file.

I'm also not convinced of the representation, as it doesn't represent a diff. I think something like

```rust
enum EofDiff {
  Addition, Deletion
}

enum FileDiff {
   Plain {
     ...
     eof: Option<EofDiff>
   }
}
```

xphoniex: I added it in `ModifiedFile` struct instead as it's the only place we're using it currently, and also simpler.

xphoniex: https://github.com/radicle-dev/radicle-surf/commit/28ce86151723f79d8d58de57050aabc8d6fe754f

xphoniex:
@FintanH take a look [here](https://app.radicle.network/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkyghsrokxzxpy9pww69xr11dr9q7edbxfo/commits/4f9c60f9ca4c023544b2625d3adb4cc40276d23a). Currently, the `lineNum` for "No newline at end of file" is the same as previous line so they get combined.

We decided to fix it here once, since there are multiple consumers for api.

FintanH:
> take a look [here](https://app.radicle.network/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkyghsrokxzxpy9pww69xr11dr9q7edbxfo/commits/4f9c60f9ca4c023544b2625d3adb4cc40276d23a). Currently, the lineNum for "No newline at end of file" is the same as previous line so they get combined.

Ah, that makes sense, thanks for the clarification.

> Re tests, it's a pretty small change & already tested locally.

Appreciate that you tested locally, but testing is also to help understand current behaviour and also ward off accidental changing of that behaviour. I'd still appreciate some testing :)

surf/src/diff.rs
```diff
@@ -88,6 +88,19 @@ pub struct CopyFile {
...
 }
 
+#[cfg_attr(
+    feature = "serialize",
+    derive(Serialize),
+    serde(rename_all = "camelCase")
+)]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum EofNewLine {
+    NoneMissing,
```
FintanH, line 98:
```suggestion
```

Since this is used with an `Option` we don't need this case. I got tipped off when I looked below and you were always calling the function with `Some` :)

FintanH (changes requested):

surf/src/diff.rs
cloudhead: Indeed, and then we could change this type to `MissingNewline { Old, New, Both }` perhaps

xphoniex: It's not always called with `Some`, example [here](https://github.com/radicle-dev/radicle-surf/pull/201/files#diff-6ff820339b0f953493d4229a5daceea07f09a7b92ed4677a89266f9174941ab8R291) and [here](https://github.com/radicle-dev/radicle-surf/pull/201/files#diff-6ff820339b0f953493d4229a5daceea07f09a7b92ed4677a89266f9174941ab8R448). Having `NoneMissing` would mean it was processed and being `None` means it was not processed, as we also need to process cases of file addition and deletion (which is not coverered here, in `ModifiedFile`).

I can change it if you still want?

kim:
> if `git2::Diff` could be somehow stored in memory?

You can create one from a textual patch: https://docs.rs/git2/0.13.25/git2/struct.Diff.html#method.from_buffer

surf/src/diff.rs
```diff
@@ -274,7 +288,7 @@ impl Diff {
                                     let mut path = parent_path.borrow().clone();
                                     path.push(new_file_name.clone());
 
-                                    diff.add_modified_file(path, vec![]);
+                                    diff.add_modified_file(path, vec![], None);
```
FintanH, line 291: Actually, thanks for pointing this out! This is dead code :) I grepped for the usage of the `diff` that calls `collect_diff` and it's only used in the `examples` code.

surf/src/diff.rs
FintanH: As I pointed out in my other comment one of those cases is essentially dead code so let's discount it. For the binary case, my feeling is that the consumer is just going to ignore `eof` anyway, so I think my point still stands. However, I'm not sure I understand your point about addition and deletion -- I thought this is only regarding modified files?

FintanH:
> You can create one from a textual patch: https://docs.rs/git2/0.13.25/git2/struct.Diff.html#method.from_buffer

Either this or we add a case in `git-platinum`. Perhaps @rudolfs and @geigerzaehler are also interested in this and how it's presented in Upstream?

xphoniex:
> > if `git2::Diff` could be somehow stored in memory?
> 
> You can create one from a textual patch: https://docs.rs/git2/0.13.25/git2/struct.Diff.html#method.from_buffer

Ah, thanks. I was looking at docs of an older version and missed this.

xphoniex:
Hm... `from_buffer` method works, but lines are missing `old_lineno` and `new_lineno` which makes `LineDiff` fail too, any experience with this @FintanH @kim ?

FintanH:
> Hm... from_buffer method works, but lines are missing old_lineno and new_lineno which makes LineDiff fail too

Haven't used this before, personally. Do you have some code I can try running on my side and help investigate?

xphoniex:
> Haven't used this before, personally. Do you have some code I can try running on my side and help investigate?

```
// surf/src/diff/git.rs
#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn test_both_missing_eof_newline() {
        let buf = r#"diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=123
\ No newline at end of file
+hello=1234
\ No newline at end of file"#;

        let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
        let diff = Diff::try_from(diff).unwrap();
        assert_eq!(diff.modified.len(), 1);
    }
}
```

surf/src/diff.rs
xphoniex: Fixed here https://github.com/radicle-dev/radicle-surf/commit/53eb72f32cf9478248d5322b8210fe20c03a4368#diff-6ff820339b0f953493d4229a5daceea07f09a7b92ed4677a89266f9174941ab8R97-R101

surf/src/diff.rs
```diff
@@ -310,7 +323,7 @@ impl Diff {
 
                                 old_entry_opt = old_iter.next();
                                 new_entry_opt = new_iter.next();
-                            },
+                            }
```
FintanH, line 326: Can we avoid these formatting changes? :)

FintanH:
@xphoniex it seems like it wants a _whole_ patch file, which you can get from `git format-patch`. I set one up in a test repo and got an example here:

```rust
        #[test]
        fn test_both_missing_eof_newline() {
            let buf = r#"
From e1f4980b5ef3cf4ae777c8b888fbd8f7a0d05876 Mon Sep 17 00:00:00 2001
From: Fintan Halpenny <fintan.halpenny@gmail.com>
Date: Fri, 18 Feb 2022 14:55:14 +0000
Subject: [PATCH] yOlO

---
 test | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/test b/test
index 29ba004..7fa0056 100644
--- a/test
+++ b/test
@@ -1 +1 @@
-hello=123
+hello=1234
-- 
2.31.1"#;

            let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
            let diff = Diff::try_from(diff).unwrap();
            assert_eq!(diff.modified.len(), 1);
        }
    }
 ```
 
 So if you could create one with/without newlines we'd be golden.

**Comment by xphoniex** [comment]
Your sample works but the second you add the culprit `"\ No newline at end of file"` to 
 either side, you'll get the error. Also you don't need the whole thing, try:
```
diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=123
+hello=1234
```
and then:
```
diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=123
\ No newline at end of file
+hello=1234
\ No newline at end of file
```

we probably could fix this by checking if content is ^ [here](https://github.com/radicle-dev/radicle-surf/blob/master/surf/src/diff/git.rs#L54) and do something with it (I didn't know what to return instead to test locally).

File: surf/src/diff.rs (hunk 4 above)
Comment (xphoniex) [reply surf/src/diff.rs:326]: Ah, they snuck in this time. Is there any particular reason we don't want them in? Cause I have to cherry pick each time I make a new change, or do you have a custom config for your fmt?

Comment (FintanH) [reply surf/src/diff.rs:326]: Ya there's a custom config here https://github.com/radicle-dev/radicle-surf/blob/master/.rustfmt.toml :)

I think you'd need to run with `cargo +nightly`

**Comment by cloudhead** [comment]
Oh nice find, I think we can work with that.

**Comment by xphoniex** [comment]
I added the tests but there's an underlying bug in `libgit2` which makes tests fail when using `git2::Diff::from_buffer` and returns the opposite side. Here's the PR I opened that would fix this, if gets merged: https://github.com/libgit2/libgit2/pull/6240.

And here's a code snippet to demonstrate the bug with a local repo:

```
$ git diff 17ec0f3 394a717
diff --git a/.env b/.env
index f9fb22f..ed7b51e 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=12345
+hello=123456
\ No newline at end of file
```

Code:

```
use git2::Repository;
//use radicle_surf as surf;

fn print_diff(diff: &git2::Diff) {
    diff.print(
        git2::DiffFormat::Patch,
        |_diff_delta, _diff_hunk, diff_line| {
            println!(
                "{} {:?} {:?}",
                diff_line.origin(),
                diff_line.origin_value(),
                std::str::from_utf8(diff_line.content()).unwrap(),
            );
            true
        },
    )
    .unwrap();
}

fn main() {
    let repo = match Repository::open("./end-git-test") {
        Ok(repo) => repo,
        Err(e) => panic!("failed to open: {}", e),
    };
    
    let hash = git2::Oid::from_str("17ec0f30306d0c80e509e1b48892cbc7c0dcc1cd").unwrap();
    let tree1 = repo.find_commit(hash).unwrap().tree().unwrap();

    let hash = git2::Oid::from_str("394a71726c3e9a86131d7160733bf9270c34deb6").unwrap();
    let tree2 = repo.find_commit(hash).unwrap().tree().unwrap();

    let diff = repo
        .diff_tree_to_tree(Some(&tree1), Some(&tree2), None)
        .unwrap();
    print_diff(&diff);
    //let rad_diff = surf::git::Diff::try_from(diff).unwrap();
    //println!("diff is {:?}", rad_diff);

    println!();
    let buf = r#"
diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=12345
+hello=123456
\ No newline at end of file
"#;
    let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
    print_diff(&diff);
    //let diff = surf::git::Diff::try_from(diff).unwrap();
    //println!("diff = {:?}", diff);
}
```

Output:
```
F FileHeader "diff --git a/.env b/.env\nindex f9fb22f..ed7b51e 100644\n--- a/.env\n+++ b/.env\n"
H HunkHeader "@@ -1 +1 @@\n"
- Deletion "hello=12345\n"
+ Addition "hello=123456"
< DeleteEOFNL "\n\\ No newline at end of file\n"

F FileHeader "diff --git a/.env b/.env\nindex f89e4c0..7c56eb7 100644\n--- a/.env\n+++ b/.env\n"
H HunkHeader "@@ -1 +1 @@\n"
- Deletion "hello=12345\n"
+ Addition "hello=123456\n"
> AddEOFNL "\\ No newline at end of file\n"
```

So the two non-working test cases are [commented out](https://github.com/radicle-dev/radicle-surf/pull/201/files#diff-38b40439b6de5f6a50a7d41d42733c2ebff150d2a0e4cadb305f0752ad261ea0R214) for now, and will be uncommented once the PR is merged.

**Comment by FintanH** [comment]
Happy to accept and merge these changes. Nice one on tracking down that libgit2 bug :smile: 

I'm just wondering if you considered my comment here https://github.com/radicle-dev/radicle-surf/pull/201#issuecomment-1047876074?

Thanks for the time and patience!

**Comment by xphoniex** [comment]
> Happy to accept and merge these changes. Nice one on tracking down that libgit2 bug smile
> 
> I'm just wondering if you considered my comment here [#201 (comment)](https://github.com/radicle-dev/radicle-surf/pull/201#issuecomment-1047876074)?
> 
> Thanks for the time and patience!

Yup, I started using `line.origin_value()` instead of `.origin()` as it's more clear in the code. 

Re putting it on the line itself, we're already sending the `type` and currently eof nl missing line would be:
```
{
  type: "deletion | addition", // <- we can change this
  line: "\n\\ No newline at end of file\n",
  lineNum: 1
}
```
which sends two other redunant fields as well. I think a flag on the file itself would be more appropriate as suggested by Alexis here too https://github.com/radicle-dev/radicle-surf/pull/201#discussion_r807034639.

FintanH (approved):
Thanks again! The lint error on CI seems to be do with compiler updates so I'll just merge and fix in another patch :)

